PR #120, opened 2022-12-15
Add storage proof verification

### LINKED ISSUE #62: Implement `verify_storage_proofs`
## Description

We need to implement the verification of the storage proof when we access StarkNet state.
To do that we must use `pathfinder_getStorageProofs `endpoint.

The PR in PathFinder is here: https://github.com/eqlabs/pathfinder/pull/726
It has been merge.

## Acceptance criteria

- verify_storage_proofs is implemented in core library
- associated unit tests are implemented
- implementation is tested against a path finder node on a testnet

CLI and Rest API Wrapper will be done in follow up PRs.

pscott:
# Pull Request type

Adds the ability to verify a storage proof (as defined [here](https://github.com/eqlabs/pathfinder/issues/714)).
Closes #62 

- [x] Feature

Can't verify proofs

Issue Number: #62 

# What is the new behavior?

Can verify proofs

- Can verify contract proofs as well as storage proofs

- [x] No

# Other information

- Added a `data.json` file for testing purposes (since RPC endpoint is not written yet, and pathfinder doesn't expose the endpoint either yet ([actually it does now](https://github.com/eqlabs/pathfinder/pull/787)).
- The `data.json` file was obtained by doing a curl on a modified version of pathfinder.
- `pathfinder` uses `starkhash` and here we use `FieldElement` so there's a bit of wrapping / translation going on.
- The verification is simply a method called on the `GetProofOutput` which should be easy to use in the future
- I have only added a `NonMembership` test because I don't have a contract to look at now. I believe this could come in a subsequent PR (and maybe create an associated issue to add more tests)

### DIFF
--- a/beerus_core/src/lightclient/starknet/mod.rs
+++ b/beerus_core/src/lightclient/starknet/mod.rs
@@ -8,6 +8,8 @@ use starknet::{
 };
 use url::Url;
 
+pub mod storage_proof;
+
 #[automock]
 #[async_trait]
 pub trait StarkNetLightClient: Send + Sync {
--- a/beerus_core/src/lightclient/starknet/storage_proof.rs
+++ b/beerus_core/src/lightclient/starknet/storage_proof.rs
@@ -3,14 +3,14 @@ use serde::{Deserialize, Deserializer};
 use starknet::core::crypto::pedersen_hash;
 use starknet::core::types::FieldElement;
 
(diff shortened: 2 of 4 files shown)


abdelhamidbakhta (approved):
Lgtm
Other tests can come in follow up PR
Thanks a lot

PR #194, opened 2023-02-01
Feat: Migrate to helios 0.2.0

danilowhk:
Update Beerus to use Beerus 0.2.0 which is wasmable.

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: N/A

# What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

# Other information

### DIFF
--- a/beerus_core/src/config.rs
+++ b/beerus_core/src/config.rs
@@ -3,6 +3,7 @@ use std::str::FromStr;
 use ethers::types::Address;
 use eyre::{eyre, Result};
 use helios::config::{checkpoints, networks::Network};
+use std::path::PathBuf;
 
 pub const DEFAULT_ETHEREUM_NETWORK: &str = "goerli";
 // By default, we use the Ethereum Mainnet value.
@@ -24,6 +25,8 @@ pub struct Config {
     pub starknet_rpc: String,
     // StarkNet core contract address.
     pub starknet_core_contract_address: Address,
+
+    pub data_dir: Option<PathBuf>,
 }
 
 impl Config {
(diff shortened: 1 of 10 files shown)


beerus_core/src/config.rs
```diff
@@ -42,13 +45,15 @@ impl Config {
         let starknet_core_contract_address = std::env::var("STARKNET_CORE_CONTRACT_ADDRESS")
             .unwrap_or_else(|_| DEFAULT_STARKNET_CORE_CONTRACT_ADDRESS.to_string());
         let starknet_core_contract_address = Address::from_str(&starknet_core_contract_address)?;
+        let data_dir = PathBuf::from("/tmp");
```
drspacemn, line 48: Is the desired result that this is set by the user? If so we need to check an environment variable is set here.

Or can default to "/tmp" if none is set.

beerus_core/tests/config.rs
```diff
@@ -245,12 +247,13 @@ mod tests {
     #[test]
     fn given_ethereum_network_is_invalid_when_ethereum_network_then_returns_error() {
         let config = Config {
-            ethereum_network: "invalid".to_string(),
+            ethereum_network: "mainnet2".to_string(),
```
drspacemn, line 250: "invalid".to_string() is more explicit here as "mainnet2" isn't obviously a wrong network name upon reading.

danilowhk: Ok!

beerus_core/src/config.rs
danilowhk: Ok! Will move to .env

drspacemn (approved):

PR #208, opened 2023-02-22
dev: workspace run top level binaries

drspacemn:
Moved workspace declarations including:
- name
- description
- husky dev dependencies
- [[bin]] 

Into the crates.

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)
- [x] Build-related changes
- [ ] Testing

must change directory into a crate in order to run

Issue Number: N/A

# What is the new behavior?

ability to run --bin <binary_name> from the top level dir

- [ ] Yes
- [x] No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -498,14 +498,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "beerus"
-version = "0.1.0"
-dependencies = [
- "beerus_core",
- "cargo-husky",
-]
-
-[[package]]
 name = "beerus-rpc"
 version = "0.1.0"
 dependencies = [
@@ -529,6 +521,7 @@ version = "0.1.0"
 dependencies = [
  "base64 0.20.0",
  "beerus_core",
+ "cargo-husky",
  "clap",
  "env_logger 0.10.0",
  "ethers",
@@ -547,6 +540,7 @@ name = "beerus_core"
 version = "0.1.0"
 dependencies = [
  "async-trait",
+ "cargo-husky",
  "ethers",
  "eyre",
  "helios",
@@ -568,6 +562,7 @@ version = "0.1.0"
 dependencies = [
  "base64 0.20.0",
  "beerus_core",
+ "cargo-husky",
  "env_logger 0.9.3",
  "ethers",
  "eyre",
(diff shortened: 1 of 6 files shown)


dlight:
What about, instead of having explicit `[[bin]]`s, to add `beerus_cli` and `beerus_rest_api` as [default-members](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-default-members-field) in the top level `Cargo.toml`? Like this:

```toml
[workspace]
members = ["beerus_core", "beerus_cli", "beerus_rest_api"]
default-members = ["beerus_cli", "beerus_rest_api"]
```

Then `cargo run --bin beerus_cli` and `cargo run --bin beerus_rest_api` works by default.

Another option is doing nothing at all and just running `cargo run --package beerus_cli` (or `cargo run -p beerus_cli`) at the top level.

drspacemn:
> What about, instead of having explicit `[[bin]]`s, to add `beerus_cli` and `beerus_rest_api` as [default-members](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-default-members-field) in the top level `Cargo.toml`? Like this:
> 
> ```toml
> [workspace]
> members = ["beerus_core", "beerus_cli", "beerus_rest_api"]
> default-members = ["beerus_cli", "beerus_rest_api"]
> ```
> 
> Then `cargo run --bin beerus_cli` and `cargo run --bin beerus_rest_api` works by default.
> 
> Another option is doing nothing at all and just running `cargo run --package beerus_cli` (or `cargo run -p beerus_cli`) at the top level.

Turns out the rework of the top level `Cargo.toml` actually fixed the `--bin` issue. So `cargo run --bin beerus_cli` works without the `[[bin]]` declaration or `default_members`.

As to your point about doing nothing. I think adding the ability to call with `--bin` is valuable to have.

dlight:
> Turns out the rework of the top level Cargo.toml actually fixed the --bin issue. So cargo run --bin beerus_cli works without the [[bin]] declaration or default_members.

Oh, nice!

dlight (approved):
I guess the `beerus` top level crate existed just to add `cargo-husky` to it, right?

In this case it looks like a good idea to get rid of it and add `cargo-husky` to workspace members, since the hook is added only once.

So, looks good to me.

PR #242, opened 2023-03-09
feat: add starknet_getNonce RPC endpoint

dpinones:
* Implemented `starknet_getNonce`
# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #211 

# What is the new behavior?

Making a request to JSON-RPC `starknet_getNonce` returns the nonce at a given address on Starknet.

- [ ] Yes
- [x] No

# Other information


beerus-rpc/src/beerus_rpc_server.rs
```diff
@@ -13,13 +16,27 @@ pub struct BeerusRpc {
...
 }
 
 #[async_trait]
 impl BeerusApiServer for BeerusRpc {
     async fn hello_world(&self) -> Result<String> {
         Ok("Hello World!".to_string())
     }
+
+    async fn stark_nonce(&self, contract_address: String) -> Result<String> {
+        let contract_address = FieldElement::from_str(&contract_address).unwrap();
```
drspacemn, line 31: As we are trying to get std out of the library what are your thoughts on defaulting to 
`FieldElement::from_hex_be` for the time being?

beerus-rpc/example_rpc_calls/stark_nonce.hurl
```diff
@@ -0,0 +1,7 @@
+POST http://0.0.0.0:3030
+Content-Type: application/json
+{
+    "jsonrpc":"2.0","method":"stark_nonce","params":[
+        "0x06ee3440b08a9c805305449ec7f7003f27e9f7e287b83610952ec36bdc5a6bae"
+    ],"id":1
+}
```
drspacemn, line 7: nit: newline here

beerus-rpc/src/beerus_rpc_server.rs
dpinones: Yes, using `FieldElement::from_hex_be` instead of `FieldElement::from_str` could be a good solution if we are certain that the input is in `hexadecimal` format, which would allow us to avoid the dependency on `std`.

If the input could be in a different format, such as `decimal`, then we could consider writing a custom conversion function that does not depend on `std`.

I think that in this case using the `FieldElement::from_hex_be` function would be sufficient.

drspacemn: Okay, lets stick to `from_hex_be` for the time being, and if this comes up as a feature request or issue later we can add the conversion

drspacemn: @danilowhk thoughts here?

dpinones: I already use the `from_hex_be` method

drspacemn (approved):
lgtm

PR #275, opened 2023-03-13
Feat/pending transactions 231

lancenonce:
# Pull Request type

- [ ] Bugfix
- [ x ] Feature
- [ ] Build-related changes
- [ ] Testing

We are migrating query_pending_transactions to beerus_rpc. 

Issue Number: #231 

# What is the new behavior?

- Created a new starknet_pending_transactions function in the rpc's server.rs
- Created a starknet_pending_transactions function in the BeerusLightClient implementation
- Defined a test under beerus.rs in the core file

- [ ] Yes
- [ x ] No

# Other information

This PR should not be merged yet because I still need assistance properly implementing the function in the BeerusLightClient implementation. Here is the code snippet:

Screenshot 2023-03-13 at 1 52 34 PM

The goal is to fetch the pending transactions from the starknet client and return them if they exist. We throw an error if none are available. I also want to ask whether or not we should throw an error with no pending transactions. Perhaps an error should only occur when the call fails?


crates/beerus-rpc/src/server.rs
```diff
@@ -83,6 +83,10 @@ trait BeerusApi {
         &self,
         tx_hash: String,
     ) -> Result<MaybePendingTransactionReceipt>;
+
+    #[method(name = "starknet_pendingTransactions")]
+    async fn starknet_pending_transactions(&self) -> Result<Vec<Transaction>>;
```
phklive, line 88: This is correct 👍

crates/beerus-rpc/src/server.rs
```diff
@@ -233,6 +237,18 @@ impl BeerusApiServer for BeerusRpc {
...
             .unwrap())
     }
+
+    async fn starknet_pending_transactions(&self) -> Result<Vec<Transaction>> {
+        let transactions = self
+            ._beerus
+            .starknet_lightclient
+            .pending_transactions()
+            .await
+            .unwrap();
```
phklive, line 247: Needs proper implementation of error handling.

crates/beerus-rpc/src/server.rs
```diff
@@ -233,6 +237,18 @@ impl BeerusApiServer for BeerusRpc {
             .await
             .unwrap())
     }
+
+    async fn starknet_pending_transactions(&self) -> Result<Vec<Transaction>> {
```
phklive, line 241: The function implementation is correct, this is how it should be done.

crates/beerus-core/tests/beerus.rs
```diff
@@ -3957,6 +3957,37 @@ mod tests {
         assert_eq!(result.unwrap_err().to_string(), expected_error.to_string());
     }
 
+    #[tokio::test]
+    async fn given_empty_result_when_calling_starknet_pending_transactions_then_should_return_error(
```
phklive, line 3961: Great that you have been adding tests, are they passing ?

crates/beerus-core/src/lightclient/beerus.rs
```diff
@@ -471,4 +471,20 @@ impl BeerusLightClient {
...
+    /// See https://github.com/starknet-io/starknet-addresses for the StarkNet core contract address on different networks.
+    /// # Arguments
+    /// # Returns
+    /// `Ok(U256)` if the operation was successful - A vector of pending transactions
+    /// `Err(eyre::Report)` if the operation failed - No pending transactions found.
+    // TODO: Determine if error should throw if no pending transactions are found, or just a zero case
+    pub async fn starknet_pending_transactions(&self) -> Result<Vec<Transaction>> {
+        let transactions = self.starknet_lightclient.pending_transactions().await;
+        
+        match transactions {
```
phklive, line 485: You cannot match here you will get a type error: 
you are only able to match on Option<> and Enum<> but here you are trying to match on a vec<Transaaction>.

You should check if the vec.len() == 0, to know if it is empty or not.

phklive: Imo if it is empty you return the empty vector, it just means that no transactions are pending, which is possibly some informations that people want to know too.

Only return an error if you have an actual error, like a bad call or others...

Wdythink @drspacemn?

phklive (changes requested):

crates/beerus-rpc/src/server.rs
lancenonce: Noted

crates/beerus-core/tests/beerus.rs
lancenonce: They aren't passing because the code doesn't compile from the implementation

crates/beerus-core/src/lightclient/beerus.rs
lancenonce: Got it! yeah that's what I was thinking. So the error case would be if the call itself fails?

lancenonce:
All tests pass, completed functionality for the pending transactions functions. Changes requested by @phklive resolved, but still need review. One question: I unwrapped the result in the rpc server, but this was after using map_err to log the Error. Will this still cause Rust to panic if an error is introduced?

Issue #226, opened 2023-03-06
feat: Migrate Starknet : get_estimate_fee from "beerus_rest_api" to "beerus_rpc"


danilowhk:
# Feature Request

**Describe the Feature Request**

We are deprecating "beerus_rest_api" and now migrating all the api endpoints to "beerus_rpc".

Migrate get_estimate_fee and add tests on beerus_rpc.

**Describe Preferred Solution**

**Describe Alternatives**

**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

dubzn:
Hi Danilo, can i try this one?

PR #344, opened 2023-03-27
Migrate starknet_estimateFee()

dubzn:
# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: [#226](https://github.com/keep-starknet-strange/beerus/issues/226)

# What is the new behavior?
Migrate the starknet_get_estimate_fee method from beerus_rest_api to beerus_rpc, which given a block_id and a broadcasted_transaction returns the estimate_fee

- [ ] Yes
- [x] No

# Other information

:warning: This PR replace the older one (https://github.com/keep-starknet-strange/beerus/pull/270)


dubzn:
@phklive also add error handling as in the other rpc methods, if you see that it is necessary to change/add something to the code, please let me know. Thankss

phklive:
Before approving PR please take example on merged code and use the correct error handling patterns.

dubzn:
> Before approving PR please take example on merged code and use the correct error handling patterns. https://github.com/keep-starknet-strange/beerus/blob/main/crates/beerus-rpc/src/lib.rs#L163

Hii, thanks for feedback. Sorry for the delay, I wrote the code based on [this method](https://github.com/keep-starknet-strange/beerus/blob/741ae7739eb60851f9cce2e6511b7e11a86ce841/crates/beerus-rpc/src/lib.rs#L194) but I see that there are others that directly do the unwrap (like the one you gave me), which one should I go with? and the same with _broadcasted_transaction.

What do you think about this impl?
```
async fn starknet_estimate_fee(
        &self,
        block_id_type: String,
        block_id: String,
        broadcasted_transaction: String,
    ) -> Result<FeeEstimate, Error> {
        let block_id = block_id_string_to_block_id_type(&block_id_type, &block_id).unwrap();

        let broadcasted_transaction: BroadcastedTransaction =
            serde_json::from_str(&broadcasted_transaction).unwrap();

       Ok(self
            .beerus
            .starknet_lightclient
            .estimate_fee(broadcasted_transaction, &block_id)
            .await
            .unwrap())  // no catch errors here?
    }
```

I have doubts about what type of error should I use, in BerusApiError there is not one assigned for the estimateFee, should I create a specific one for this case or reuse any of these? Why in some cases do you use the mapping to an error and in others do you unwrap directly? sorry for the questions

PR #368, opened 2023-04-08
feat: Migrate Starknet: call from "beerus_rest_api" to "beerus_rpc" 

dubzn:
# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #361 

# What is the new behavior?

- Add **call** endpoint for beerus_rpc

- [ ] Yes
- [x] No

# Other information


crates/beerus-rpc/src/lib.rs
```diff
@@ -429,4 +430,18 @@ impl BeerusApiServer for BeerusRpc {
...
+        &self,
+        request: FunctionCall,
+        block_number: u64,
+    ) -> Result<Vec<FieldElement>, Error> {
+        let call_result = self
+            .beerus
+            .starknet_lightclient
+            .call(request, block_number)
+            .await
+            .map_err(|e| Error::Call(CallError::Failed(anyhow::anyhow!(e.to_string()))))?;
```
phklive, line 444: Incorrect error handling, correct way would be: 
`.await.map(|_| Error::from(BeerusApiError::X))`

examples/beerus-rpc/starknet_call.hurl
```diff
@@ -0,0 +1,15 @@
+POST http://0.0.0.0:3030
+Content-Type: application/json
+{
+  "jsonrpc":"2.0",
+  "method":"starknet_call",
+  "params": {
+    "request": {
+      "contract_address": "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
+      "entry_point_selector": "0x361458367e696363fbcc70777d07ebbd2394e89fd0adcaf147faccd1d294d60",
```
phklive, line 9: Well done here

crates/beerus-rpc/tests/rpc.rs
```diff
@@ -205,4 +204,27 @@ mod tests {
...
+    async fn starknet_call() {
+        let beerus_rpc = setup_beerus_rpc().await;
+
+        let request = FunctionCall {
+            contract_address: FieldElement::from_hex_be(
+                "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
+            )
+            .unwrap(),
+            entry_point_selector: FieldElement::from_hex_be(
+                "0x361458367e696363fbcc70777d07ebbd2394e89fd0adcaf147faccd1d294d60",
```
phklive, line 218: lgtm

crates/beerus-rpc/src/api.rs
```diff
@@ -211,4 +212,11 @@ pub trait BeerusApi {
         block_id: String,
         broadcasted_transaction: String,
     ) -> Result<FeeEstimate, Error>;
+
+    #[method(name = "call")]
+    async fn call(
+        &self,
```
phklive, line 218: lgtm

phklive (changes requested):
Update error handling by taking example on other implemented methods, rest is correct

crates/beerus-rpc/src/lib.rs
dubzn: Thanks for the feedback, I have a question about what type of error I should use from BerusApiError. And another thing that makes me doubt is if it should bring the error message sent by the client, if I send a generic error like `.map_err(|_| Error::from(BeerusApiError::InternalServerError))`, it would be unclear what is the reason of the failure, right?

With InternalApiError, we will always get this message.
image

Older one:
image
image
image

I think it would be convenient to put some optional argument to be able to pass the error message through the different layers, what do you think?

phklive: Interesting, I indeed think it could be important to pass that was returned as message.

What do you think @drspacemn, should we stay strict as papyrus and keep our errors and handle them manually or should we use the returned errors directly ?

phklive:
Hey man, how is it going?

phklive (approved):
lgtm

Issue #392, opened 2023-04-13
feat: Update starknet_getStorageAt


phklive:
# Feature Request

**Describe the Feature Request**

On beerus_core and then on the rpc starknet_getStorageAt has been wrongly implemented using the latest block and not taking BlockId as a parameter. Fix this by adding back BlockId as a parameter in the method.

**Describe Preferred Solution**

**Describe Alternatives**

**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

phklive:
Still open

phklive:
Let's go forward on this @dpinones

PR #398, opened 2023-04-14
feat: update starknet_get_storage_at

dpinones:
update `starknet_get_storage_at`

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #392 

# What is the new behavior?

In `beerus_core` and `beerus_rpc`, the `starknet_getStorageAt` method receives a `BlockId` as a parameter.

- [ ] Yes
- [x] No

# Other information


crates/beerus-core/src/lightclient/beerus.rs
```diff
@@ -282,35 +282,6 @@ impl BeerusLightClient {
...
-    /// # Arguments
-    ///
-    /// * `contract_address` - The StarkNet contract address.
-    /// * `storage_key` - The storage key.
-    ///
-    /// # Returns
-    ///
-    /// `Ok(FieldElement)` if the operation was successful.
-    /// `Err(eyre::Report)` if the operation failed.
-    pub async fn starknet_get_storage_at(
```
dpinones, line 297: This file contains all the methods that interact with StarkNet and Ethereum. However, the `get_storage_at` method now only interacts with StarkNet. It makes sense to remove it from here and use the one from `starknet_lightclient`. I  also remove the integration tests from `beerus_core` since they don't make sense anymore.
@phklive  what do you think?

phklive: We need to keep using this one, using the last block from ethereum makes sure that the data that we return is valid and safe. 

If not a starknet node could lie considering that the data would not be validated on L1 yet.

crates/beerus-core/src/lightclient/beerus.rs
```diff
@@ -282,35 +282,6 @@ impl BeerusLightClient {
...
-        &self,
-        contract_address: FieldElement,
-        storage_key: FieldElement,
-    ) -> Result<FieldElement> {
-        let last_block = self
-            .ethereum_lightclient
-            .read()
-            .await
```
phklive, line 305: Please add this code back

phklive (commented):
Let's keep the pr very lean, just update starknet_get_storage_at to support blockId

crates/beerus-core/src/lightclient/beerus.rs
dpinones: I already reverted the changes.

crates/beerus-core/tests/integration.rs
```diff
@@ -17,60 +17,6 @@ mod test {
...
-    #[tokio::test]
-    async fn given_ethereum_light_client_returns_error_when_starknet_get_storage_at_should_fail() {
-        let server = MockServer::start();
-        let mock_request = mock_get_storage_at(&server);
-        let config = mock_server_config(&server);
-
-        let starknet_lightclient = Box::new(StarkNetLightClientImpl::new(&config).unwrap());
-        let mut helios_lightclient = MockEthereumLightClient::new();
-        let expected_error = "Ethereum light client error";
-
```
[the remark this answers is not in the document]
dpinones: I already reverted the changes.

crates/beerus-core/src/lightclient/beerus.rs
```diff
@@ -298,17 +298,36 @@ impl BeerusLightClient {
...
+            .await?
+        {
+            if block.block_number <= last_proven_block {
+                return self
+                    .starknet_lightclient
+                    .get_storage_at(contract_address, storage_key, block_id)
+                    .await;
+            }
+        }
+        Err(eyre::eyre!("BlockId is not proven yet"))
```
dpinones, line 330: Here I am ensuring that the `blockId` received as a parameter is validated by L1 and we are also following the StarkNet Specs. What do you think @phklive

glihm: From what we have talked @dpinones, looks like we are doing too much validations that are also done by the full node under our feet.

I think @phklive had other idea in mind with this comment:

    We need to keep using this one, using the last block from ethereum makes sure that the data that we return is valid and safe. 
    If not a starknet node could lie considering that the data would not be validated on L1 yet.
    
Or @phklive what do you think? We should keep the way @dpinones implemented it?

Issue #383, opened 2023-04-13
bug: wasm32 target broken (helios related)


glihm:
# Bug Report

**Beerus version:**

Last version, using the new head of helios.

**Current behavior:**

Can't build wasm32 target due to helios issue (https://github.com/a16z/helios/issues/226).

**Expected behavior:**

Compiled to wasm32 ok.

**Steps to reproduce:**

Build wasm32 with:
`cargo build -p beerus-core --no-default-features --target wasm32-unknown-unknown --release`

**Related code:**

```
insert short code snippets here
```

**Other information:**

PR #400, opened 2023-04-15
fix: change helios revision with a version that compiles

glihm:
# Pull Request type

- [X] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

CI is broken by helios dependency. This PR changes the revision of helios to have a version that
compiles fine in wasm32 and allows us to pass all tests in the CI for wasm32.

Howerver, it's important to note that some work still pending on helios for wasm32 target.
But at least, beerus is working fine in wasm32, only call to `eth_call` will fail in wasm32.

Issue Number: #383

# What is the new behavior?

Now wasm32 target compiles fine.

-
-
-

- [ ] Yes
- [X] No

# Other information

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ edition = "2021"
 beerus-core = { path = "./crates/beerus-core" }
 
 # Helios revision is supposed to be changed by the corresponding tag once the revision is released. Next expected tag: 0.5.0.
-helios = { git = "https://github.com/a16z/helios", rev = "f63e4142c8c8094cdf94cf8d1ef1cb637a7eac98" }
+helios = { git = "https://github.com/glihm/helios", rev = "688ed805a142ef6ead361889cc6d42a4cdede918" }
 
 #TODO: update with version tag instead of hash rev when available
 starknet = { git = "https://github.com/xJonathanLEI/starknet-rs", rev = "e72cc0a4b3ea67c7691ebec58300da3c21f0abb2" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -883,7 +883,7 @@ dependencies = [
 [[package]]
 name = "client"
 version = "0.4.0"
-source = "git+https://github.com/a16z/helios?rev=f63e4142c8c8094cdf94cf8d1ef1cb637a7eac98#f63e4142c8c8094cdf94cf8d1ef1cb637a7eac98"
+source = "git+https://github.com/glihm/helios?rev=688ed805a142ef6ead361889cc6d42a4cdede918#688ed805a142ef6ead361889cc6d42a4cdede918"
 dependencies = [
  "common",
  "config",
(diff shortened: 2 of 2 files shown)


glihm:
I am not sure how the error on `mainnet` and `goerli` is generated by this commit. Any idea?

glihm:
after a full cleanup and retest, I have this behavior:

1. Fist time the test command is issued after the build, it fails.
2. Second time it's run without any modification, it passes.

Is there something related to env variables behavior during the tests?

dpinones:
> I am not sure how the error on `mainnet` and `goerli` is generated by this commit. Any idea?

Hey @glihm , yesterday I got this error in my PR. I reran the job and the test passed correctly. However, I couldn't find why this error was generated.

glihm:
Thank you for the feedback @dpinones and thank you for running the test again. Seems all good.

I am thinking about the `thread=2` in the testing command line. I don't know how rust is working with this, but we have no locks, and several tests that are playing with env variables. Maybe we can remove the `thread=2` for now, to ensure we are not introducing races in the test suite.

dpinones:
> Thank you for the feedback @dpinones and thank you for running the test again. Seems all good.
> 
> I am thinking about the `thread=2` in the testing command line. I don't know how rust is working with this, but we have no locks, and several tests that are playing with env variables. Maybe we can remove the `thread=2` for now, to ensure we are not introducing races in the test suite.

@glihm  worked on this in pr #401

Cargo.toml
```diff
@@ -17,7 +17,7 @@ edition = "2021"
 beerus-core = { path = "./crates/beerus-core" }
 
 # Helios revision is supposed to be changed by the corresponding tag once the revision is released. Next expected tag: 0.5.0.
-helios = { git = "https://github.com/a16z/helios", rev = "f63e4142c8c8094cdf94cf8d1ef1cb637a7eac98" }
+helios = { git = "https://github.com/glihm/helios", rev = "688ed805a142ef6ead361889cc6d42a4cdede918" }
```
phklive, line 20: I see that you are pointing directly at a version in your personal github profile, what is the plan with that and porting back to main helios repo?

phklive (commented):
Explanation needed.

phklive:
How is going the discussion with Helios to make sure that it will get fixed on their side ?

Cargo.toml
glihm: The plan is that to have wasm32 PR accepted i must check with the maintainer how to fix an issue related to threading in wasm32. This may take a long time so it means if we wait this pr to be merged... we may have during a long time a CI that will fail.

The idea is that I'll keep my repo updated with Helios main, and we will get back to Helios main when the wasm32 proble is fixed.

glihm:
It's not an easy fix and the maintainer is also busy in lots of subjects. So with my little knowledge it will take time to fix this as I have lots of tests and researches to do. That's why I proposed to change to my fork to avoid having a CI failing over and over until it's fixed.

Because in my fork it's compiling fine, only a runtime issue to fix.

phklive:
@drspacemn what is your opinion about pointing to @glihm's repo?

glihm:
I'll close this PR as it's not the current focus of the work to be done. And maybe re-open it when wasm32 will be more in the scope.

phklive:
Nice move, let's organize and tackle that in due time.

Issue #397, opened 2023-04-14
feat: Add proper error handling on starknet rpc endpoints


phklive:
# Feature Request

**Describe the Feature Request**

We were previously figuring out the way to handle rpc errors on Beerus hence incorrect implementation on endpoints implemented before we managed to settle for a specific way to handle ( we took example on pathfinder error handling ) errors as seen here:
- [Pathfinder error struct](https://github.com/eqlabs/pathfinder/blob/a4f08758862cdfee5fed5ec093305d8b1bea78d2/crates/rpc/src/error.rs#L9)
- [Beerus error struct](https://github.com/keep-starknet-strange/beerus/blob/main/crates/beerus-rpc/src/api.rs#L21)

The mission is to update all rpc endpoints to support correct error handling, a good understanding of how the rpc works and all the methods the rpc implements is required considering you will need to understand what errors are returned by each method and in what situation. ( remove unwraps, remove anyhow, etc... ) We must prevent beerus from crashing at all costs.

**Describe Preferred Solution**

**Describe Alternatives**

**Related Code**

**Additional Context**

Here are some additional ressources that could help: 
- [Beerus error handling in method](https://github.com/keep-starknet-strange/beerus/blob/main/crates/beerus-rpc/src/lib.rs#L79)

You can also take inspiration on Pathfinder:
- [Pathfinder repo](https://github.com/eqlabs/pathfinder)

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

dubzn:
I would like to take this one

phklive:
assigned.

phklive:
How are you doing with this one?

dubzn:
Hey Paul, how you doing? I have this [PR](https://github.com/keep-starknet-strange/beerus/pull/402) ready for the review, and in case that is ok I implemented the error handling in a similar way in [this one](https://github.com/keep-starknet-strange/beerus/pull/408) (I wanted to validate that's why still in draft, actually, there is no work to be done in that)

PR #402, opened 2023-04-16
Refactor starknet_getEstimateFee + getEstimateFee error handling

dubzn:
# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Testing

Issue Number: #397 

# What is the new behavior?
- starknet_estimateFee method now takes BlockId & BroadcastedTransaction parameters instead of BlockId and String.
- starknet_estimateFee now can throw BLOCK_NOT_FOUND, CONTRACT_NOT_FOUND and CONTRACT_ERROR errors via BeerusApiError

- Change starknet_estimateFee param from String to BroadcastedTransaction
- Add error handling for starknet_estimateFee (proposal for handler multiple types of errors), as described in the Starknet standard, estimateFee can throw the following [errors](https://github.com/starkware-libs/starknet-specs/blob/e0b76ed0d8d8eba405e182371f9edac8b2bcbc5a/api/starknet_api_openrpc.json#L483):
image

- [X] Yes
- [ ] No

Changing the method param of starknet_estimateFee
# Other information


crates/beerus-rpc/src/lib.rs
```diff
@@ -372,32 +374,31 @@ impl BeerusApiServer for BeerusRpc {
...
             .await
-            .map_err(|e| Error::Call(CallError::Failed(anyhow::anyhow!(e.to_string()))))?;
-        Ok(estimate_fee)
+            .map_err(|e| {
+                let error_type: String = e.to_string();
+                match error_type.as_str() {
+                    BLOCK_NOT_FOUND => Error::from(BeerusApiError::BlockNotFound),
+                    CONTRACT_ERROR => Error::from(BeerusApiError::ContractError),
+                    CONTRACT_NOT_FOUND => Error::from(BeerusApiError::ContractNotFound),
+                    _ => Error::from(BeerusApiError::ContractError),
```
dubzn, line 399: Let me know what you think of this, I could replicate it in methods that throw more than one error, but the way to do it is open to change.

crates/beerus-rpc/src/api.rs
```diff
@@ -16,6 +17,10 @@ use starknet::{
         Transaction,
     },
 };
+use std::string::ToString;
+pub const BLOCK_NOT_FOUND: &str = "Block not found";
+pub const CONTRACT_NOT_FOUND: &str = "Contract not found";
+pub const CONTRACT_ERROR: &str = "Contract error";
```
dubzn, line 23: If we go this way I think it would be ideal to separate these constants in another file.

glihm:
Synced with @dubzn to check this. Here a brief recap:

1. We have seen that lot's of work is done in `starknet-rs` about typings and codegen from the spec. So @dubzn will propose a rework of this PR based on those observations to stick to the generated code that have high chance to become the source of truth for all Rust-based implementation of starknet projects.

2. Even if the JSON-RPC spec of Starknet has a very defined set of errors, we have a situation. The implementation **MUST** deal with invalid value. For instance, if the error code sent back by the node in the response is `500000`, we can't map that to an actual value of the specification. For this reason, dub will explore a way to correctly handle this.

3. Finally, for ETH endpoints, we will remove any abstraction possible to work with helios types as much as possible, to ensure that future users of Beerus (that may use helios today) can find the exact same API.

4. Dub will probably integrate this PR in the other PR more generic with all endpoints.
   @dubzn will focus on only on Starknet endpoints.

@phklive @drspacemn this is the debrief from @dubzn and I after syncing today on this. :+1:

phklive:
Okay great, let's see what that translates in in terms of code and let's go 👍

drspacemn (approved):
lgtm

crates/beerus-rpc/src/api.rs
drspacemn: good for now, can separate at a later date

crates/beerus-rpc/src/lib.rs
drspacemn: good for now

Issue #366, opened 2023-04-07
bug: starknet query-transaction-by-hash fails because of cli options


glihm:
# Bug Report

**Beerus version:**

branch `main`, commit `1e423ba32a90fcd9e637f16e1d535d208a35f34f`.

**Current behavior:**

When I try to run: `beerus starknet query-transaction-by-hash -h 0x938439759375...` the CLI
is complaining about option name collision.

**Expected behavior:**

The hash can be passed without generating this error.

**Steps to reproduce:**

Compile beerus on the mentioned commit `cargo build --all [--release]`, and then run `.\target\debug\beerus.exe starknet query-transaction-by-hash -h 0x938439759375....`.
The error also happens without giving `-h`.

**Related code:**

I still not very rust fluent, I'll double check from where this can be fixed.
I just wanted to open this issue to not forget fixing it.

**Other information:**

Error from rust:
```
thread 'main' panicked at 'Command query-transaction-by-hash: Short option names must be unique for each argument, but '-h' is in use by both 'hash' and 'help' (call `cmd.disable_help_flag(true)` to remove the auto-generated `--help`)', C:\...\.cargo\registry\src\github.com-1ecc6299db9ec823\clap-4.1.13\src\builder\debug_asserts.rs:115:17
```

glihm:
This issue seems to be observed only for `debug` builds. I have no problem running the release buid:
`.\target\release\beerus.exe starknet query-transaction-by-hash --hash 0x06a....af --config C:\tmp\beerus_main.toml`

glihm:
I can take that one to find the reason for this to be able to work in debug as well.

dpinones:
According to the [documentation](https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes), if no value is assigned to the `short` attribute, it takes the first character of the field name by default. This is causing a conflict for us. I propose removing all `short` attributes.

glihm:
Seems to be a good idea to start @dpinones. At least we avoid surprises when adding options.

I am still wondering why this issue is only visible in debug, perhaps the help is added only on debug? But I'm not sure.

phklive:
Closed because of removal of beerus-cli

PR #406, opened 2023-04-18
fix: remove short attribute to avoid collision

dpinones:
* remove `short` attribute to avoid option name collision
# Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #366 

# What is the new behavior?

Beerus CLI works correctly and does not throw the option collision error.

- [ ] Yes
- [x] No

# Other information

### DIFF
--- a/crates/beerus-cli/src/model.rs
+++ b/crates/beerus-cli/src/model.rs
@@ -53,18 +53,18 @@ pub enum EthereumSubCommands {
     /// Sends a Raw Transaction.
     SendRawTransaction {
         /// Bytes of the Raw Transaction
-        #[arg(short, long, value_name = "BYTES")]
+        #[arg(long, value_name = "BYTES")]
         bytes: String,
     },
     /// Query the balance of an Ethereum address.
     QueryBalance {
         /// The address to query the balance of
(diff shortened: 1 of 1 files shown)


phklive (commented):
What type of "collision" did the short attribute add?

dpinones:
> 
Basically, if you run a CLI command that receives parameters, it doesn't work. According to the `clap` documentation, if no value is assigned to the `short` attribute, it takes the first character of the field name by default. This is causing a collision for us. That's why I removed the short attribute.
Moreover, using the short attribute in this way does not seem appropriate for the ux. We should create a discussion to ensure that our CLI has the best ux.

phklive:
The CLI will be rebuilt as one of the next steps before release of V1 to improve UX and use what we have improved for the RPC. 

This PR is interesting but not needed at this moment.

We should focus for now on finalizing the tests for the RPC.

phklive:
Just tried: 

cargo r --bin beerus starknet query-nonce --address xxx

On main and it worked, can you test please.

phklive:
The current focus is unaligned with PR and CLI might be deprecated, if not this pr will be merged.

Issue #385, opened 2023-04-13
feat: Dockerize Beerus


phklive:
# Feature Request

**Describe the Feature Request**

Run beerus in a Docker container and enable it to run with github actions.

**Describe Preferred Solution**

**Describe Alternatives**

**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

robinstraub:
I can work on this alongside #387 with @glihm

dpinones:
assigned

glihm:
Ready to work with you @robinstraub. 👍🏻

PR #416, opened 2023-04-22
test: add assertions to hurl files

robinstraub:
This PR adds a hurl execution script 

# Pull Request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Documentation content changes
- [x] Testing
- [ ] Other (please describe):

# What is the current behavior?

Currently hurl files have to run one by one.

Issue Number: #387 

# What is the new behavior?

Hurls can be executed with a single script.

# Does this introduce a breaking change?

- [ ] Yes
- [x] No

# Other information

PR #421, opened 2023-04-23
build: dockerize beerus

robinstraub:
This pull request provides a docker configuration for running Beerus locally and in the CI

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [x] Build-related changes
- [ ] Testing

Beerus currently needs to be run from source or through beerusup. No docker setup also complicates running end-to-end testings in the CI

Issue Number: #385 

# What is the new behavior?

A "Dockerfile" file allows Beerus to be built as a docker container.

There is also a docker-compose.yml to simplify running the docker container (e.g. setting up port-forwarding / environment variables. This docker-compose.yml can serve as a basis to later add devnets so we don't need to rely on external nodes and calling the mainnet.

An extra workflow is also added to demonstrate how to build the docker image and execute it in the CI.

- [ ] Yes
- [x] No

# Other information

**image build time (local)**
Image first build takes ~10min 
Image second build takes ~1min

**image build time (ci)**
Image first build takes ~20min
Image second build takes ~5min

Co-authored with @glihm 🎉


robinstraub:
Looks like the helios port is not properly open or forwarded. The following command:

```
curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' --retry 6 --retry-delay 10 --retry-connrefused http://0.0.0.0:3031
```

Returns an error.

```
curl: (52) Empty reply from server
```

robinstraub:
Needs PR #416 to be merged for this to be ready (hurl tests are failing without it)

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder
+COPY --from=planner /beerus/recipe.json recipe.json
+RUN cargo chef cook --release --recipe-path recipe.json
+COPY . .
+RUN cargo build --all --locked --release
+
+FROM debian:buster-slim AS runtime
```
robinstraub, line 14: We could replace the debian base image by an alpine to get a lighter image, however this would require either to disable cargo chef layer caching, or restricting the target to x86 linux. In the context of the CI I kept a heavier version in order to allow:
- faster builds with cache (25min -> 5min)
- support to build beerus for other architecture (e.g. arm)

glihm:
@robinstraub as @Aragar199 fixed the CPU problem, we should be able to move forward with this mate? What's your opinion on this?

phklive (approved):
lgtm

Issue #422, opened 2023-04-26
dev: Hurl tests should assert valid results


robinstraub:
We have a hurl file for every endpoint implemented in Beerus. Some hurl files are failing to get a result -- calling them return an error.

The CI currently asserts for these errors : this is a bad behavior, every hurl file should run a request that returns a valid result, and this result should be correctly asserted.

Here is a listing of the Hurl files that provide bad assertions:

| Method failing with hurl tests        | Error                                                                                                                                                              | 
|:--------------------------------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| starknet_addDeclareTransaction        | `{"code":-32602,"message":"No more params"}`                                                                                                                       |
 | starknet_addDeployAccountTransaction  | {"code":-32602,"message":"invalid type: string \"[0x1234, 0x5678]\", expected a sequence at line 2 column 26"}                                                     |
| starknet_addInvokeTransaction         | `{"code":-32602,"message":"No more params"}`                                                                                                                       |
| starknet_getContractStorageProof      | `{"code":-32602,"message":"invalid type: string \"0x0206F38F7E4F15E87567361213C28F235CCCDAA1D7FD34C9DB1DFE9489C6A091\", expected a sequence at line 2 column 72"}` |
 | eth_blockNumber                       | malformatted number                                                                                                                                                |
 | eth_call                              | `{"code":40,"message":"Contract error"}`                                                                                                                           |
 | eth_chainId                           | malformatted number                                                                                                                                                |
 | eth_estimateGas                       | malformatted number                                                                                                                                                |
 | eth_gasPrice                          | malformatted number                                                                                                                                                |
 | eth_getBalance                        | malformatted number                                                                                                                                                |
 | eth_getBlockTransactionCountByNumber  | malformatted number                                                                                                                                                |
| eth_getCode                           | `{"code":40,"message":"Contract error"}`                                                                                                                           |
 | eth_getStorageAt                      | `{"code":-32602,"message":"invalid length 1, expected a (both 0x-prefixed or not) hex string or byte array containing 32 bytes at line 2 column 13"}`              |
| eth_getTransactionByBlockHashAndIndex | `{"code":40,"message":"Contract error"}`                                                                                                                           |
 | eth_getTransactionCount               | malformatted number                                                                                                                                                |
 | eth_maxPriorityFeePerGas              | malformatted number                                                                                                                                                |
[...]

PR #423, opened 2023-04-27
fix malformatted number to fit eth hex format (#422)

glihm:
# Pull Request type

- [X] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

`eth_` endpoints returning numbers are displayed in decimal. 

Issue Number: #422 

# What is the new behavior?

`eth_` endpoints returning numbers are now displayed as hexadecimal string.

- [ ] Yes
- [X] No

# Other information

This PR is not conflicted with #416 and can be merged to achieve some issues listed in #422.

### DIFF
--- a/crates/beerus-rpc/src/lib.rs
+++ b/crates/beerus-rpc/src/lib.rs
@@ -70,7 +70,7 @@ impl BeerusRpcServer for BeerusRpc {
             .await
             .map_err(|_| Error::from(BeerusApiError::ContractError))?;
 
-        Ok(utils::u256_to_hex_string(balance))
+        Ok(hex_string!(balance))
     }
 
     async fn eth_get_transaction_count(
@@ -90,7 +90,7 @@ impl BeerusRpcServer for BeerusRpc {
             .await
             .map_err(|_| Error::from(BeerusApiError::ContractError))?;
 
-        Ok(utils::u64_to_hex_string(tx_count))
+        Ok(hex_string!(tx_count))
     }
 
     async fn eth_get_block_transaction_count_by_hash(&self, hash: &str) -> Result<String, Error> {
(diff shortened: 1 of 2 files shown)


crates/beerus-rpc/src/lib.rs
```diff
@@ -181,7 +184,7 @@ impl BeerusRpcServer for BeerusRpc {
             .get_gas_price()
             .await
             .map_err(|_| Error::from(BeerusApiError::ContractError))?;
-        Ok(gas_price.to_string())
+        Ok(format!("0x{gas_price:x}"))
```
phklive, line 187: We need to check if all methods that should return hex considering the Ethereum rpc specs: 

Indeed return hex, and does that should not, do not, so we are 100% compliant.

crates/beerus-rpc/src/lib.rs
```diff
@@ -69,7 +69,7 @@ impl BeerusRpcServer for BeerusRpc {
             .await
             .map_err(|_| Error::from(BeerusApiError::ContractError))?;
 
-        Ok(balance.to_string())
+        Ok(format!("0x{balance:x}"))
```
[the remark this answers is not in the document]
glihm: I started doing like this, but I didn't want to add a utils crate. I can add a module utils in EPC, which yes you are right should be more organized.

crates/beerus-rpc/src/utils.rs
```diff
@@ -0,0 +1,11 @@
+use ethers::types::U256;
```
phklive, line 1: I think that the utils file should go in the beerus-core utils folder: 
https://github.com/keep-starknet-strange/beerus/blob/main/crates/beerus-core/src/ethers_helper.rs

phklive (commented):
Potential modification of placement of utils fn. 

Rest lgtm.

crates/beerus-rpc/src/utils.rs
glihm: It's a very generic function of formatting we may use it for other RPC purposes.

What if tomorrow some starknet output must be hexadecimal?

But if it's sound better to you, let me know and I'll move it to the place you point out.

glihm: Or perhaps changing to a generic function or macro.

PR #436, opened 2023-05-05
feat: fix typo helios_client state_root selector

joshualyguessennd:
# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing
- [x] Other (please describe):

A typo error in the definition of a variable selector

Issue Number: N/A

# What is the new behavior?

- [ ] Yes
- [x] No

# Other information


crates/beerus-core/src/lightclient/ethereum/helios_lightclient.rs
```diff
@@ -176,7 +176,7 @@ impl EthereumLightClient for HeliosLightClient {
         // Corresponds to the StarkNet core contract function `stateRoot`.
         // The function signature is `stateRoot() -> (uint256)`.
         // The function selector is `0x95d8ecA2`.
-        let data = vec![0x95, 0x88, 0xec, 0xa2];
+        let data = vec![0x95, 0xd8, 0xec, 0xa2];
```
glihm, line 179: Can you provide @joshualyguessennd a bit more context about this? Thanks a lot for this catch.

joshualyguessennd: the comment defines the selector function as 0x95d8ecA2 while the vector does not match vec![0x95, 0x88, 0xec, 0xa2]; the typo that I have just seen comes from the selector defined in the comment

glihm: Thanks a lot @joshualyguessennd and sorry for the delay.
The problem here is the comment, not the code. That's a nice catch.

The `keccak256` of `stateRoot()` which is the function of the starknet core contract called is `9588eca2b322322b87d370149dfa69fc4982323abc8166a070bd09d96309087f`.

And we only take the 4 first bytes -> which are correct in the vector. The typo can be fixed in the comment btw. :)

glihm:
Thanks for the correction and sorry for the delay.
@phklive @drspacemn can be merged to avoid future confusion. :+1:

phklive (approved):
lgtm

Issue #438, opened 2023-05-06
bug: Beerus stops processing incoming connections on hosts with less than 3 cpu


robinstraub:
# Bug Report

**Beerus version:**

 v0.2.0

**Current behavior:**

*edit: this issue only happens in a Docker container*

When running Beerus on a host with 2 logical cpus, under some specific circumstances (some method calls sent at a specific time), Beerus stops reacting to incoming connections, without logging error, and without panicking. Subsequent RPC calls thus result in timeouts. 

When the cpu count gets lower (1 or fraction of a cpu), Beerus fails to synchronize the state root and the last proven block (in what I refer to as "[the loop](https://github.com/keep-starknet-strange/beerus/blob/main/crates/beerus-core/src/lightclient/beerus.rs#L136)"). This behavior also leads to Beerus stopping to react to incoming connections as with 2 cpus.

**Expected behavior:**

As a light client, Beerus should not require a host with 3 logical CPUs to operate.

**Steps to reproduce:**

1. Execute Beerus on a host with 2 logical cpus or less
2. Call a Beerus RPC method that interacts with the starknet core contract (e.g. `starknet_l1_to_l2_message_nonce`), while Beerus synchronization task (the loop) is calling it (the task calls it [here](https://github.com/keep-starknet-strange/beerus/blob/main/crates/beerus-core/src/lightclient/beerus.rs#L147))

The easiest way to achieve this is to run Beerus inside a Docker container so the CPU can be restricted (`docker run --cpus=2 [...]`), you can find a working docker configuration (Dockerfile and docker-compose.yml) on [PR 421](https://github.com/keep-starknet-strange/beerus/pull/421).

**Related code:**

N/A

**Other information:**

This behavior has been a bit erratic and Beerus randomly worked with 2 cpu. After some investigation, when RPC keeps working on 2CPU (e.g. allowing all hurl tests to be conducted without timeout), this happens because Beerus task (the loop) panics. Here is a log:

```
beerus-beerus-rpc-1  | thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: out of sync: 12 slots behind
beerus-beerus-rpc-1  | 
beerus-beerus-rpc-1  | Location:
[...]

PR #440, opened 2023-05-09
fix: Beerus stops processing incoming connections on hosts with less than 3 cpu

Aragar199:
bug: Beerus stops processing incoming connections on hosts with less than 3 cpu

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Beerus stops processing incoming connections on hosts with less than 3 cpu due to read threads being created that are locking resources and starving on low cpu environments.

Issue Number: #438 

# What is the new behavior?

Removed read threads from block syncing loop that were locking up resources.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/beerus-core/src/lightclient/beerus.rs
+++ b/crates/beerus-core/src/lightclient/beerus.rs
@@ -126,7 +126,6 @@ impl BeerusLightClient {
             self.starknet_lightclient.start().await?;
             self.sync_status = SyncStatus::Synced;
 
-            let ethereum_clone = self.ethereum_lightclient.clone();
             let starknet_clone = self.starknet_lightclient.clone();
             let node_clone = self.node.clone();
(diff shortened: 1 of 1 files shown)


crates/beerus-core/src/lightclient/beerus.rs
```diff
@@ -126,32 +126,13 @@ impl BeerusLightClient {
...
 
-            let ethereum_clone = self.ethereum_lightclient.clone();
             let starknet_clone = self.starknet_lightclient.clone();
             let node_clone = self.node.clone();
             let poll_interval_secs = self.config.get_poll_interval();
 
             // Define function that will loop
             let task = async move {
                 loop {
-                    let state_root = ethereum_clone
```
Aragar199, line 137: These are locking threads and starving the app when run on low cpu environments.

PR #441, opened 2023-05-11
fix: Beerus stops processing incoming connections on hosts with less than 3 cpu

Aragar199:
bug: Beerus stops processing incoming connections on hosts with less than 3 cpu

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Beerus stops processing incoming connections on hosts with less than 3 cpu. Readers are created that starve the thread.

Issue Number: #438 

# What is the new behavior?

readers are capped to prevent deadlock when running on env with 2 cpu

- [ ] Yes
- [x] No

### DIFF
--- a/crates/beerus-core/src/lightclient/beerus.rs
+++ b/crates/beerus-core/src/lightclient/beerus.rs
@@ -6,7 +6,7 @@ use gloo_timers::callback::Interval;
 #[cfg(not(feature = "std"))]
 use wasm_bindgen_futures::spawn_local;
 
-use tokio::sync::RwLock;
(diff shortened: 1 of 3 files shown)


robinstraub:
Can confirm this allows tests to pass on Docker and thus on the CI

glihm:
that's a huge work, very nice guys!
I have tested locally on the dockers I have spawn for the tests, all green. :rocket:

@phklive @drspacemn what do you think?

phklive (approved):
lgtm

PR #483, opened 2023-05-17
fix: eth_getStorageAt

dpinones:
* Update `examples/beerus-rpc/eth/eth_getStorageAt.hurl`
# Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #471 

# What is the new behavior?

This method passes the end-to-end testing CI of Beerus.

- [ ] Yes
- [x] No

# Other information

Screenshot from 2023-05-16 22-36-25


dpinones:
With the update to the RPC example, it should now pass the CI.

glihm:
Nice catch @dpinones. And yes basically the `ethereum_type` is enforcing `H256` to have exactly all the required digits to be a valid hexadecimal hash.

Which means, as you said, even if UX is a bit less appreciable, we have to pass the full string. If we were working with `U256`, the convertion from `"0x0"` would have been valid though.

Last thing that your PR and comment make me think about is that, we have said we try to limit at the maximum the abstraction we put on the top of helios. And here basically we are forced to add a validation step, because we rely on helios as a library, strongly typed due to Rust.

So there is no choice here, if we want to call helios functions, we must at least parse arguments from string to ensure we have the good type for underlying helios client methods. :+1:

phklive (approved):
lgtm

PR #490, opened 2023-06-01
RPC error handling refactor

dubzn:
# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

In several RPC methods (referring to Starknet), they directly use .unwrap(), which would cause the application to crash if something goes wrong. Therefore, the goal is to eliminate the .unwrap() calls and also allow identifying the error that occurred in the JSON-RPC response (with its code and message).

Issue Number: #397 

# What is the new behavior?

### The following modifications were made:

- A modification was made to the /lib.rs file of the RPC, removing the .unwrap() calls.
- BeerusApiError was completely refactored to be able to take error codes and custom messages from a JsonRpcError.
- The Rust docs of all modified methods were updated.
- Tests were modified and fixed based on the changes made in beerus-cores.

The Starknet Light Client was modified as follows:
- The Eyre dependency was removed (debatable, but during the implementation, it caused more problems because the error code couldn't be retrieved with the existing code).
- The client methods now return the JsonRpcError type of error, which provides flexibility to understand the error code and message. This allows transforming the error case into a BeerusApiError type error (without losing the original error message for transparency).

### Here are a few concerns that I have, and I would like to hear a second opinion:

- What do you think about the error conversion across layers: JsonRpcClientError (jsonrpc module) -> JsonRpcError (lightclient) -> BeerusApiError (beerus-rpc/api) -> Error (beerus-rpc/lib)?
- I decided to remove Eyre from the client because, from that point on, we couldn't access the error code. This was because errors were constructed from a string, making error handling difficult from that point onwards.

Mainly in the lightclient/beerus.rs file:

- Since we now return errors of type JsonRpcError, in several places where errors occur, I decided to use an error with code 520 because there wasn't a perfect match for the problem. On the positive side, if it fails, the error message will be transparent and declarative when it reaches the JSON-RPC response.

To make these changes, I relied on starknet-rs for the messages, codes, and overall implementation structure.

- [ ] Yes
- [x] No

# Other information
The errors will now be displayed in the following way:
image


dubzn:
One could say that the proposal is ready; I would like you guys to take a good look at it and give me your opinion. There some things that need to be adjusted (read PR description please). 

The tests are broken due to the changes, but I wanted to show you the progress and start fixing the tests tomorrow.

crates/beerus-rpc/tests/rpc.rs
```diff
@@ -33,21 +32,22 @@ mod tests {
...
+    //         .unwrap_err();
+    //
+    //     let beerus_rpc_err = ErrorObjectOwned::from(err);
+    //     let block_not_found = BeerusApiError::from(BeerusApiError::BlockNotFound as i64);
+    //     assert_eq!(beerus_rpc_err.code(), block_not_found as i8);
+    //     assert_eq!(
+    //         beerus_rpc_err.message(),
+    //         block_not_found.to_string()
+    //     );
+    // }
```
dubzn, line 50: This test was not working or covering that particular case as it was failing due to a JSON parsing error. I think we need to improve how we mock the data when conducting the tests. Additionally, we should add RPC tests covering the relevant cases now that we can identify them with error codes and messages.

drspacemn:
Taking some time to look at it. I am also worried about this:
```
What do you think about the error conversion across layers: JsonRpcClientError (jsonrpc module) -> JsonRpcError (lightclient) -> BeerusApiError (beerus-rpc/api) -> Error (beerus-rpc/lib)?
```

Did you think of any potential solutions?

PR #546, opened 2023-10-27
dev: Refactor Beerus

drspacemn:
An attempt to simplify the Beerus Light Client AND implement storage proofs in the API


crates/cli/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[async_std::main]
+async fn main() {
+    // env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
+    let subscriber = FmtSubscriber::builder().with_max_level(Level::INFO).finish();
+
+    tracing::subscriber::set_global_default(subscriber).expect("setting default subscriber failed");
+
+    let args = Args::parse();
+    let config = Config::from_file(&args.conf);
```
tonypony220, line 23: todo from env?

crates/core/src/client.rs
```diff
@@ -0,0 +1,253 @@
...
+
+impl BeerusClient {
+    /// Create a new Beerus Light Client service.
+    pub async fn new(config: Config) -> Self {
+        #[cfg(not(target_arch = "wasm32"))]
+        let mut helios_client: Client<FileDB> = config.to_helios_client().await;
+        #[cfg(target_arch = "wasm32")]
+        let mut helios_client: Client<ConfigDB> = config.to_helios_client().await;
+
+        helios_client.start().await.expect("could not init helios client");
```
tonypony220, line 97: helios doesn't fail if the endpoints are incorrect like it did; So that now beerus throws panics from a thread in a loop. Maybe consider adding some request and exit right away in case of a bad configuration

crates/core/src/config.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn from_file(path: &str) -> Self {
+        let raw_conf = fs::read_to_string(path).unwrap_or_else(|_| panic!("unable to read file: {path}"));
+
+        if path.contains(".toml") {
+            toml::from_str(&raw_conf).unwrap()
+        } else if path.contains(".json") {
+            serde_json::from_str(&raw_conf).unwrap()
+        } else {
+            Self::default()
```
tonypony220, line 104: a bit tricky

README.md
```diff
@@ -19,215 +19,52 @@
...
 
-```bash
-cp examples/.env.example .env
-source .env
-```
-
-#### Examples
 ```bash
 cargo run -p beerus-core --example basic
 ```
```
tonypony220, line 70: todo or del this line `BEERUS_CONFIG=path/to/config.toml cargo run -p beerus-core --example basic`?

crates/rpc/src/api.rs
```diff
@@ -0,0 +1,418 @@
...
+        let l1_block_num = self.beerus.get_local_block_id(block_id).await;
+        self.beerus.starknet_client.estimate_message_fee(message, l1_block_num).await.map_err(BeerusRpcError::from)
+    }
+
+    async fn block_number(&self) -> Result<u64, BeerusRpcError> {
+        Ok(self.beerus.get_local_block_num().await)
+    }
+
+    async fn block_hash_and_number(&self) -> Result<BlockHashAndNumber, BeerusRpcError> {
+        Ok(BlockHashAndNumber { block_hash: FieldElement::ZERO, block_number: self.beerus.get_local_block_num().await })
```
tonypony220, line 292: `FieldElement::ZERO` is stub?

crates/cli/src/main.rs
drspacemn: Good point, added a env option of the config flag isn't set. 
You think that UX is okay/understandable?

crates/core/src/config.rs
drspacemn: tricky i.e. shouldn't set default if there isn't any config file?

Am good w/ panicing if there isn't a good config where the user pointed

README.md
drspacemn: Would you want to add to this PR and get the examples running w/ the new layout?

crates/core/src/config.rs
tonypony220: Yeah. Better panic because it’s explicit. 
As a bad idea, it can be simply `--conf` `path_to_file` or `default` or `env`

crates/rpc/src/api.rs
drspacemn: thanks for the catch, will change

crates/cli/src/main.rs
tonypony220: If you look at the code - absolutely!
But if you blindly run it like `cargo run -p beerus-cli` without any `ENV` var , and see panic like 
```
thread 'main' panicked at /mount/beerus_new/crates/core/src/config.rs:163:14:
incorrect helios client config: execution client creation error: relative URL without a base
```
it may confuse imho

README.md
tonypony220: Sure

crates/cli/src/main.rs
drspacemn: what do you think would be a better way to handle?

Display the CLI usage if there is no config and no ENV vars set?

crates/core/src/config.rs
drspacemn: I like the idea to panic here. That way if they have old env vars set and init w/ a conf path they don't end up with unexpected state.

README.md
tonypony220: While doing examples I started to think that maybe users shoud be abstracted from beerus core like rpc? 
I've created dummy PR to show what I mean. [Link](https://github.com/keep-starknet-strange/beerus/pull/547)

crates/rpc/src/api.rs
```diff
@@ -0,0 +1,420 @@
...
+        block_id: BlockId,
+        contract_address: FieldElement,
+        keys: Vec<FieldElement>,
+    ) -> Result<StorageProof, BeerusRpcError>;
+
+    #[method(name = "provenStateRoot")]
+    async fn proven_state_root(&self) -> Result<FieldElement, BeerusRpcError>;
+
+    #[method(name = "provenBlockNumber")]
+    async fn proven_block_number(&self) -> Result<u64, BeerusRpcError>;
```
tonypony220, line 160: It looks like the difference between `block_number()` and `proven_block_number()` is Pool seconds and additional node call, i mean maybe it's not necessary endpoint?

crates/core/src/client.rs
```diff
@@ -0,0 +1,252 @@
...
+        // allow for historical block ids
+        match provided_block_id {
+            BlockId::Number(num) => {
+                if local_block_num > num {
+                    provided_block_id
+                } else {
+                    BlockId::Number(local_block_num)
+                }
+            }
+            _ => BlockId::Number(local_block_num),
```
tonypony220, line 195: Can arg `BlockId` be hash for older block then local?

drspacemn: It can. I think we need to implement the local cache of block data first as a first big ticket after this merge then add that as an API on top of it

crates/rpc/src/api.rs
drspacemn: yup will delete

crates/cli/src/main.rs
drspacemn: look for config flag

crates/core/src/client.rs
drspacemn: To be handled in:

https://github.com/keep-starknet-strange/beerus/issues/550

examples/rpc/starknet_getBlockNumber.hurl
```diff
@@ -2,12 +2,11 @@ POST http://0.0.0.0:3030
 Content-Type: application/json
 {
     "jsonrpc": "2.0",
-    "method": "eth_chainId",
+    "method": "starknet_getBlockNumber",
```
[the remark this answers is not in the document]
drspacemn: removed

crates/core/src/client.rs
drspacemn: Added a short curcuit before we hit the loop:
```
        if let Err(err) = sn_state_root_inner(&helios_client, config.get_core_contract_address()).await {
            panic!("execution client err: \n\n{err:#?}");
        }
```

crates/rpc/src/api.rs
```diff
@@ -0,0 +1,410 @@
...
+    ) -> Result<FieldElement, BeerusRpcError> {
+        let l1_block_num = self.beerus.get_local_block_id(block_id).await;
+        let l1_root = self.beerus.get_local_root().await;
+
+        let fetched_val = self.beerus.starknet_client.get_storage_at(contract_address, key, l1_block_num).await?;
+        let mut proof = self
+            .beerus
+            .get_contract_storage_proof(l1_block_num, contract_address.as_ref(), &[*key.as_ref()])
+            .await
+            .unwrap();
```
tonypony220, line 200: empty reply

PR #547, opened 2023-10-31
dev: pub abstraction of beerusClient 

tonypony220:
https://github.com/keep-starknet-strange/beerus/pull/546#discussion_r1377656628


crates/core/src/client.rs
```diff
@@ -69,11 +69,11 @@ impl Default for NodeData {
...
     /// Ethereum light client
     #[cfg(not(target_arch = "wasm32"))]
-    pub helios_client: Arc<Client<FileDB>>,
+    helios_client: Arc<Client<FileDB>>,
     #[cfg(target_arch = "wasm32")]
-    pub helios_client: Arc<Client<ConfigDB>>,
+    helios_client: Arc<Client<ConfigDB>>,
     /// StarkNet light client
-    pub starknet_client: JsonRpcClient<HttpTransport>,
+    pub(crate) starknet_client: JsonRpcClient<HttpTransport>,
```
drspacemn, line 76: Is this just for external clarity?

tonypony220: as for now, beerus protect calling with more fresh block n then beerus local, so exposing client would be not consistent, as far as i understand

PR #597, opened 2024-02-27
test: Chain-agnostic test suite

pierre-l:
Re-implement the simplest Hurl-based test scenarios in rust

Create a new workflow for these tests: rust-integration-tests
Rename the old workflow to hurl-integration-tests

The new test suite only covers 11 scenarios out of 25 for now.

This new workflow still isn't as reliable as I'd want it to be. From what I can tell, Helios just fail to synchronize sometimes. It's nothing new. This will be my next area of focus.

### DIFF
--- a/.github/workflows/rust-integration-tests.yml
+++ b/.github/workflows/rust-integration-tests.yml
@@ -1,0 +1,33 @@
+name: rust-integration-tests
+on:
+  schedule:
+    - cron: "0 8,20 * * *"
+  push:
+  workflow_dispatch:
+
+jobs:
+  integration:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+
+      - name: build beerus
+        run: cargo build -p beerus-cli --release
+
+      - name: start beerus RPC on MAINNET
+        env:
+          BEERUS_VERSION: "ci"
+          NETWORK: "mainnet"
+          ETH_EXECUTION_RPC: ${{ secrets.ETH_EXECUTION_RPC_MAINNET }}
+          STARKNET_RPC: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET }}
+        run: ./target/release/beerus & while ! timeout 1 bash -c "echo > /dev/tcp/localhost/3030" 2> /dev/null; do sleep 2; done
+
+      - name: run integration test on MAINNET
+        run: cargo test --package beerus-rpc --features integration-tests
+
+      - name: stop RPC
+        run: kill $(lsof -t -i:3030)
+ 
--- a/crates/rpc/Cargo.toml
+++ b/crates/rpc/Cargo.toml
@@ -4,11 +4,15 @@ edition = "2021"
 name = "beerus-rpc"
 version = "0.4.0"
 
+[features]
+integration-tests = ["pretty_assertions"]
+
 [dependencies]
 beerus-core = { path = "../core" }
 starknet.workspace = true
 eyre.workspace = true
 jsonrpsee = { version = "0.20.3", features = ["macros", "server", "server-core"] }
+pretty_assertions = { version = "1.4.0", optional = true }
 serde = { workspace = true, features = ["derive"] }
 serde_with.workspace = true
 thiserror.workspace = true
@@ -16,3 +20,5 @@ tracing.workspace = true
 
 [dev-dependencies]
 serde_json = "1.0"
+reqwest = "0.11.13"
+tokio.workspace = true
--- a/crates/rpc/Cargo.toml
+++ b/crates/rpc/Cargo.toml
@@ -5,10 +5,11 @@ name = "beerus-rpc"
 version = "0.4.0"
 
 [features]
-integration-tests = ["pretty_assertions"]
+integration-tests = ["pretty_assertions", "cached"]
 
 [dependencies]
 beerus-core = { path = "../core" }
+cached = { version = "0.49.3", optional = true }
 starknet.workspace = true
 eyre.workspace = true
 jsonrpsee = { version = "0.20.3", features = ["macros", "server", "server-core"] }
@@ -19,6 +20,7 @@ thiserror.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
+lazy_static = "1.4.0"
 serde_json = "1.0"
 reqwest = "0.11.13"
 tokio.workspace = true
--- a/crates/rpc/tests/rpc.rs
+++ b/crates/rpc/tests/rpc.rs
@@ -2,16 +2,26 @@
 #![cfg(feature = "integration-tests")]
 
 use beerus_core::config::DEFAULT_PORT;
+use cached::Cached;
+use std::{ops::Deref, sync::Mutex};
+
+use cached::SizedCache;
 use reqwest::Url;
 use starknet::{
     core::types::{
-        BlockId, BlockTag, BlockWithTxHashes, BlockWithTxs, DeclareTransaction,
+        BlockId, BlockWithTxHashes, BlockWithTxs, DeclareTransaction,
         DeployAccountTransaction, FieldElement, InvokeTransaction,
         MaybePendingBlockWithTxHashes, MaybePendingBlockWithTxs,
         MaybePendingTransactionReceipt, Transaction, TransactionReceipt,
     },
-    providers::{jsonrpc::HttpTransport, JsonRpcClient, Provider},
+    providers::{
+        jsonrpc::HttpTransport, JsonRpcClient, Provider, ProviderError,
+    },
 };
+
+lazy_static::lazy_static! {
+    static ref BLOCK_CACHE: Mutex<SizedCache<u64, MaybePendingBlockWithTxs>> = Mutex::new(SizedCache::with_size(1000));
+}
 
 fn rpc_client() -> JsonRpcClient<HttpTransport> {
     let rpc_url: Url = format!("http://localhost:{}", DEFAULT_PORT)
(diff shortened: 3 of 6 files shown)


crates/rpc/Cargo.toml
```diff
@@ -17,3 +20,4 @@ tracing.workspace = true
 [dev-dependencies]
 serde_json = "1.0"
 reqwest = "0.11.13"
+tokio = { version = "1", features = ["sync", "macros", "rt", "time"] }
```
sergey-melnychuk, line 23: why not use workspace dependency here? features seem to match

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,488 @@
...
+            .transactions
+            .last()
+            .as_ref()
+            .expect("We need a last transaction here")
+            .transaction_hash(),
+    )
+    .await;
+}
+
+/* TODO
```
sergey-melnychuk, line 467: seems like a good candidate to be extracted as an issue :)

.github/workflows/rust-integration-tests.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: build beerus
+        run: cargo build -p beerus-cli --release
+
+      - name: start beerus RPC on MAINNET
+        env:
+          BEERUS_VERSION: "ci"
+          NETWORK: "mainnet"
+          ETH_EXECUTION_RPC: ${{ secrets.ETH_EXECUTION_RPC_MAINNET }}
+          STARKNET_RPC: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET }}
+        run: ./target/release/beerus &
```
sergey-melnychuk, line 26: the readiness probe might be a solution for the race condition (e.g. when tests start when Beerus is not yet up); or a fixture that ensures that Beerus is up before each test (not necessary in the same PR though)

the quick & dirty way from integration-test.sh: https://github.com/eigerco/beerus/blob/main/scripts/integration-test.sh#L12

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,488 @@
...
+                        address: None,
+                        keys: None,
+                    },
+                    last_token,
+                    page_size,
+                )
+                .await?;
+            last_token = events_page.continuation_token;
+
+            if last_token.is_none() {
```
sergey-melnychuk, line 150: `more_to_fetch` is always equal to `last_token.is_some()`, it can be substituted then such check is not necessary

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,488 @@
...
+        }
+
+        Ok(events)
+    }
+}
+
+#[tokio::test]
+async fn test_block_number() {
+    let TestContext { client, block, block_id: _, extracted_value: () } =
+        latest_block_context().await;
```
sergey-melnychuk, line 173: maybe I am missing something, but caching the fixtures seems to make sense here - otherwise *one every test* the blocks are fetched from the external RPC

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,488 @@
...
+        block_id: BlockId,
+        transaction_index: usize,
+        expected: &BlockWithTxs,
+    ) {
+        let transaction = client.get_transaction_by_block_id_and_index(block_id, transaction_index as u64).await.expect("Failed to retrieve a specific transaction by its block ID and index");
+
+        // `starknet-rs` doesn't implement `PartialEq` on its DTOs, and transactions have many *variants which makes pure Rust comparison painful.
+        // Just serialize these to compare them.
+        assert_eq!(
+            serde_json::to_string(&transaction).unwrap(),
```
sergey-melnychuk, line 383: no need to `serde_json::to_string`, as the `serde_json::Value` is `Eq` and also can deal with fields ordering

we might also want to use [`pretty_assertions`](https://docs.rs/pretty_assertions/latest/pretty_assertions/) for complex values for better readability in case test fails

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,488 @@
...
+   starknet_getStateUpdate
+   starknet_getStorateAt
+   starknet_getTransactionReceipt
+   starknet_getTransactionStatus
+   starknet_syncing
+   pathfinder_getProof
+*/
+
+fn truncate_felt_to_u128(felt: &FieldElement) -> u128 {
+    u128::from_be_bytes(felt.to_bytes_be().split_at(16).1.try_into().unwrap())
```
sergey-melnychuk, line 487:
```suggestion
    u128::from_be_bytes(felt.to_bytes_be()[16..])
```

crates/rpc/tests/rpc.rs
pierre-l: Ah! The choice for `to_string` was a poor man's `pretty_assertions` :) Thanks!

.github/workflows/rust-integration-tests.yml
pierre-l: Yes that's was I was looking into, but for some reason I didn't think of using this solution. It should work!

LKozlowski: I second that! Readiness probes are great solution for that. I always used [this script](https://github.com/vishnubob/wait-for-it) for that kind of checks.

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,483 @@
...
+    extracted_value: T,
+}
+
+/// Creates a `TestContext` with the latest block.
+async fn latest_block_context() -> TestContext<()> {
+    context(|_block| Some(())).await
+}
+
+/// Creates a `TestContext` with the latest block with at least 10 transactions.
+async fn latest_10_tx_context() -> TestContext<()> {
```
LKozlowski, line 31: nit: maybe we could use transactions count as a parameter, so we can find a block with any number or transactions if we ever need more or less, or is there a reason it is 10?

pierre-l: 10 is arbitrary. Checking all the transactions can take a long time and result in thousands of calls, so this limit is purely practical. But you are right on the name & parameter, this is inelegant.

crates/rpc/tests/rpc.rs
pierre-l: Added as a separate commit as I'm not 100% convinced this is needed/useful right now, but this should get more and more relevant as we go.

.github/workflows/rust-integration-tests.yml
sergey-melnychuk: Simple `GET /ready` handler should do here, as the server is started after Helios is synced up to a checkpoint.

pierre-l: I'm currently testing this [on my dedicated fork.](https://github.com/pierre-l/beerus-ci-test/actions). (because scheduled runs only work on the default branch)
I'll set this thread as resolved if it passes >95% of the time overnight.

crates/rpc/tests/rpc.rs
pierre-l: Will do once merged

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,477 @@
...
+struct TestContext<T> {
+    client: TestClient,
+    block: BlockWithTxs,
+    block_id: BlockId,
+    extracted_value: T,
+}
+
+/// Creates a `TestContext` with the latest block.
+async fn latest_block_context() -> TestContext<()> {
+    context(|_block| Some(())).await
```
sergey-melnychuk, line 30: If I understand correctly, what `latest_block_context` is doing is pulling latest block from the RPC and then drops it to return `()` as the extracted value. So the RPC call can be dropped, right?

Won't having two fixtures be simpler (just a question, not a suggestion)? Then each test case can use whichever is necessary and no redundant RPC calls are made:
- `with_client` that returns `TestClient`
- `with_client_and_block` that returns `(TestClient, BlockWithTxs, BlockId)`

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,477 @@
...
+            return TestContext {
+                client,
+                block_id: BlockId::Number(block.block_number),
+                block,
+                extracted_value,
+            };
+        }
+    }
+
+    panic!("No suitable block found")
```
sergey-melnychuk, line 122: Explicit `panic!` always freaks me out a bit :)

Another alternative would be returning wrapped result here (`Result` or `Option`), that can be simply unwrapped in the test code (thus pointing to a specific line & test where the problem occurred, not to this line).

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,477 @@
...
+}
+
+/// Creates a `TestContext` with the latest block.
+async fn latest_block_context() -> TestContext<()> {
+    context(|_block| Some(())).await
+}
+
+/// Creates a `TestContext` with the latest block with at least N transactions.
+async fn n_txs_context(min_tx_number: usize) -> TestContext<()> {
+    context(|block| {
```
sergey-melnychuk, line 35: Just as an alternative way of caching: here might be the least intrusive point to cache the result if necessary (maybe with [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html)).

crates/rpc/tests/rpc.rs
pierre-l: I get the reluctance to use panics, but I see more boilerplate for not additional value here.

crates/rpc/tests/rpc.rs
pierre-l: I separated the creation of the RPC client from the test context and got rid of the `test_block_number` scenario as it's redundant indeed.
[Relevant push](https://github.com/eigerco/beerus/compare/66420d6accb4502a53cdd67fab9ea9a0822c81ed..65fcee33d079b56d8e3bc88a1d51a219c05f6f1d).

crates/rpc/tests/rpc.rs
pierre-l: This does look more complicated to me ^^'

crates/rpc/tests/rpc.rs
```diff
@@ -0,0 +1,474 @@
...
+fn rpc_client() -> JsonRpcClient<HttpTransport> {
+    let rpc_url: Url = format!("http://localhost:{}", DEFAULT_PORT)
+        .parse()
+        .expect("Invalid RPC URL");
+    JsonRpcClient::new(HttpTransport::new(rpc_url))
+}
+
+struct TestClient {
+    inner: JsonRpcClient<HttpTransport>,
+    block_cache: SizedCache<u64, MaybePendingBlockWithTxs>,
```
LKozlowski, line 31: I'm trying to understand this cache mechanism and don't know how it's working, so please correct me if I'm wrong 😄 

In each test case, the Test Client (with a fresh cache) is created via the context function, and then the cache is populated in this context call. Since the context call is the only place the cache is touched, and the cache key is always different because the key gets decreased by one in a loop, and then the client is destroyed for each test. How does it work, and what is the benefit of caching here?

sergey-melnychuk: Nice catch @LKozlowski , it seems like the scope of cache is a single test-case. I believe a test-suite-wise cache would be more efficient in terms of reducing number of (duplicated) RPC calls.

But as long as those are tests, I think it is totally fine to take a step back here.
@pierre-l please feel free to drop the caching thing, no one is pushing for it really. We can totally leave it for the next PR.

sergey-melnychuk (approved):
please drop caching and let's move on with this one

Issue #599, opened 2024-02-28
feat: Migrate to the v0.6 OpenRPC spec


pierre-l:
### Motivation
Depending on an outdated version of `starknet-rs` is already becoming painful. Upgrading to the latest version should make #596 and #585 easier.
With the Starknet OpenRPC v0.7 spec coming soon, I expect v0.5.1 will be phased out.
Migrating to the Starknet OpenRPC v0.6 and `starknet-rs:0.9.0` is the way to go.

### How to
1. Make a list of v0.6-compatible providers and include that list in the README.md. Mention that Beerus will target the 0.6 version of the OpenRPC spec.
2. Restore the upgrade to v0.6. This was done a few weeks ago, but may necessitate adjustments due to the recent changes.
3. Update the tests. The tests being written for #597 might be easier to migrate.

pierre-l:
Some API providers that appear to support v0.6:

[BlastAPI](https://blastapi.io/public-api/starknet): https://starknet-mainnet.public.blastapi.io/rpc/v0_6
[Alchemy](https://docs.alchemy.com/reference/starknet-api-faq#what-versions-of-starknet-api-are-supported): https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0.6/{apiKey}
[Chainstack](https://docs.chainstack.com/reference/getting-started-starknet#starknet-json-rpc-version-endpoints): `To send a request to a specific version, append your Chainstack Starknet endpoint with the version URL suffix. For example, /rpc/v0_6.`
[Nethermind](https://docs.data.voyager.online/spec/)

PR #600, opened 2024-02-28
Bring support for the Starknet OpenRPC v0.6.0 spec

pierre-l:
Beerus is currently implicitly tied to the Starknet OpenRPC v0.5.1 spec. This PR upgrades the requirement to v0.6.0 and makes it more explicit in the README.md file. 

This is a breaking change. I bumped the Beerus version to v0.4.0 in this PR.

See #599 for more details about the motivation.

Commits will be squashed before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,39 +16,47 @@
 
 ## News
 
+* 29/02/24: Migrate to the [Starknet v0.6.0 OpenRPC spec](https://github.com/starkware-libs/starknet-specs/tree/v0.6.0), release Beerus v0.4.0
 * 17/01/24: [Eiger is taking over Beerus!](https://www.eiger.co/blog/eiger-taking-over-ownership-for-beerus-working-on-starknet-light-clients)
 
 ## Getting Started
 
-```bash
-cargo build --release
+### Configuration
 
-# insert valid api keys
-./target/release/beerus -c etc/conf/beerus.json
+Beerus relies on TWO untrusted RPC endpoints, one for L1 (Ethereum), and one for L2 (Starknet). 
+As these are untrusted they will typically not be nodes run on your local host or your local network.
 
-# wait for server to start
-hurl etc/rpc/starknet_getStateRoot.hurl
-```
+Beerus requires the [v0.6.0 of the Starknet OpenRPC specs](https://github.com/starkware-libs/starknet-specs/tree/v0.6.0).
 
-### Config
+These untrusted RPC providers must adhere to both the L1 `eth_getProof` endpoint
+as well as the L2 `pathfinder_getProof` endpoint ([detailed instructions needed!](https://github.com/eigerco/beerus/issues/602)).For this we recommend using
+[Alchemy](https://docs.alchemy.com/reference/starknet-api-faq#what-versions-of-starknet-api-are-supported) as your untrusted L2 node provider. 
 
-Beerus relies on TWO untrusted RPC endpoints. As these are untrusted they will
-typically not be nodes run on your local host or your local network. These 
-untrusted RPC providers must adhere to both the L1 `eth_getProof` endpoint
-as well as the L2 `pathfinder_getProof` endpoint. For this we recommend using
-[Alchemy](https://www.alchemy.com) as your untrusted L2 node provider. 
+More API providers can be found [here](https://docs.starknet.io/documentation/tools/api-services/).
 
-*NOTE: we rely on helios for both valid checkpoint values and consensus rpc urls*
+*NOTE: we rely on [helios](https://github.com/a16z/helios) for both valid checkpoint values and consensus rpc urls*
 
 | field   | example | description |
 | ----------- | ----------- | ----------- |
 | network | MAINNET or GOERLI | network to query |
 | eth_execution_rpc | https://eth-mainnet.g.alchemy.com/v2/YOURAPIKEY | untrusted l1 node provider url |
-| starknet_rpc | https://starknet-mainnet.g.alchemy.com/v2/YOURAPIKEY | untrusted l2 node provider url |
+| starknet_rpc | https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0.6/YOURAPIKEY | untrusted l2 node provider url |
 | data_dir | tmp | `OPTIONAL` location to store both l1 and l2 data |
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -16,40 +16,48 @@
 
 ## News
 
+* XX/XX/24: Migrate to the [Starknet v0.6.0 OpenRPC spec](https://github.com/starkware-libs/starknet-specs/tree/v0.6.0), release Beerus v0.4.0
```
pierre-l, line 19: TODO: Write the proper date before merging the PR.

sergey-melnychuk:
Nitpick: 'feat(rpc/0.6.0): ' prefix for commit messages that are related to code format is confusing ('style: ' prefix seems like a better option).

We should probably at some point add a chapter about conventional commits to the contributors guide. And then start following them ourselves :)

https://www.conventionalcommits.org/en/v1.0.0/#summary

P.S. No action item is necessary for this comment, just using it as a reminder about updating the contributors guide.

sergey-melnychuk:
> > We can't merge it until 0.6.0 is supported on mainnet, right?
> > 
> > Current design inherently enables only a single RPC-API version to be supported, which is a limitation we're enforcing by ourselves.
> > 
> > I think we might want to revisit it and maybe come up with multiple RPC versions supported simultaneously, like Pathfinder does.
> 
> @sergey-melnychuk 
> It turns out I was wrong about Alchemy not supporting RPCv0.6.0, [it just isn't the default](https://docs.alchemy.com/reference/starknet-api-faq#what-versions-of-starknet-api-are-supported).
> 
> Currently, someone trying Beerus out has no instruction on this regard and will only succeed in running it by coincidence: 0.5.1 is the most widely used version right now. Beerus is already tied to a specific version, it just isn't explicit.
> 
> Working on supporting multiple versions, or staying on 0.5.1 will slow us down on pretty much everything else.
> 
> Unrelated: I just ran the integration tests against this branch and thankfully they do pass: https://github.com/eigerco/beerus/actions/runs/8080470463/job/22077024285

@pierre-l 

Thank you for checking, we do need an explicit version check on startup in order to avoid weird integration issues.

Also my bad, 0.6.0 is supported on mainnet already, I was probably thinking 0.7.x :)

Maybe not now, just to avoid slowing things down, but we will eventually need a way to support multiple spec versions. I have a few ideas on how to make it happen while keeping the necessary efforts reasonable, I will submit them as issues later.

README.md
```diff
@@ -74,13 +82,6 @@ docker build . -t beerus
 docker run -e NETWORK=<arg> -e ETH_EXECUTION_RPC=<arg> -e STARKNET_RPC=<arg> -it beerus
 ```
 
-#### Examples
```
sergey-melnychuk, line 77: Why remove the examples section?

crates/rpc/src/error.rs
```diff
@@ -2,83 +2,215 @@ use beerus_core::CoreError;
...
-                        },
-                        MaybeUnknownErrorCode::Unknown(unknown) => unknown as i32,
-                    };
-                        ErrorObjectOwned::owned(code, message, None::<()>)
-                    }
-                    _ => ErrorObjectOwned::owned(
-                        -32601,
-                        format!("{e}"),
+            BeerusRpcError::Provider(provider_err) => match provider_err {
+                // TODO Do we actually need to own these?
```
sergey-melnychuk, line 18: Is any action necessary here before merging?

README.md
```diff
@@ -16,40 +16,48 @@
...
-| starknet_rpc | https://starknet-mainnet.g.alchemy.com/v2/YOURAPIKEY | untrusted l2 node provider url |
+| starknet_rpc | https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0.6/YOURAPIKEY | untrusted l2 node provider url |
 | data_dir | tmp | `OPTIONAL` location to store both l1 and l2 data |
 | poll_secs | 5 | `OPTIONAL` seconds to wait for querying sn state |
 | rpc_addr | 127.0.0.1:3030 | `OPTIONAL` local address to listen for rpc reqs |
 | fee_token_addr | 0x049d36...e004dc7 | `OPTIONAL` fee token to check for `getBalance` |
 
+### Running Beerus for the first time
+
+Edit the configuration file at `etc/conf/beerus.toml` and set up the API provider URLs there.
```
sergey-melnychuk, line 51: Editing a copy of the file might be a more reasonable suggestion, as the file is under git, and any accidentally committed changes might leak sensitive information (API keys), e.g. via pull request.

README.md
```diff
@@ -16,40 +16,48 @@
...
+Beerus relies on TWO untrusted RPC endpoints, one for L1 (Ethereum), and one for L2 (Starknet). 
+As these are untrusted they will typically not be nodes run on your local host or your local network.
 
-### Config
+Beerus requires the [v0.6.0 of the Starknet OpenRPC specs](https://github.com/starkware-libs/starknet-specs/tree/v0.6.0).
 
-Beerus relies on TWO untrusted RPC endpoints. As these are untrusted they will
-typically not be nodes run on your local host or your local network. These 
-untrusted RPC providers must adhere to both the L1 `eth_getProof` endpoint
+These untrusted RPC providers must adhere to both the L1 `eth_getProof` endpoint
```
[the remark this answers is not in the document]
sergey-melnychuk: Maybe add a link to the issue here

sergey-melnychuk (approved):
LGTM % a few nitpicks

pierre-l:
> Also my bad, 0.6.0 is supported on mainnet already, I was probably thinking 0.7.x :)

You probably got that wrong info from me..

> we will eventually need a way to support multiple spec versions

I'm with you here on the idea, but for the sake of minimizing the maintenance costs I think we should wait as long as it's reasonable to do so before we do implement it.
This topic gave me a couple more ideas. How about an extra coffee break tomorrow to discuss all of that?

README.md
pierre-l: Hm, I mistakenly believe these were obsolete. They are not.

crates/rpc/src/error.rs
pierre-l: Not really. I see allocations that could have been avoided here, but this isn't trivial nor needed.
Won't fix.

PR #601, opened 2024-02-29
feat: stateless tx execution

### LINKED ISSUE #573: feat: RPC codegen based on OpenRPC spec
To consider: https://github.com/sergey-melnychuk/iamgroot

### LINKED ISSUE #576: tests: make extended integration test suite
Current one is based on `hurl` when Rust code might be more expressive and flexible for it.

sergey-melnychuk:
Allow stateless execution, estimation & tracing of transactions (and calls) using [blockifier](https://github.com/starkware-libs/blockifier).

Act as a proxy for the query methods (yet verify results with `getProof` where applicable).

See also: [README](https://github.com/eigerco/beerus/blob/sm/stateless/crates/experimental-api/README.md).

Closes #573 
Closes #576


crates/api/Cargo.toml
pierre-l, line 1: I'm confused.
If this is also a server, but doesn't import Helios, then it's not a replacement for the current implementation of Beerus, is it?
Is this a proxy layer to add in front of Beerus?

crates/api/README.md
pierre-l, line 1: Please explain what this is and how to use it in addition to how to generate it.

crates/api/src/exe/err.rs
```diff
@@ -0,0 +1,61 @@
...
+pub enum Error {
+    Io(std::io::Error),
+    Reqwest(reqwest::Error),
+    IamGroot(iamgroot::jsonrpc::Error),
+    StarknetApi(starknet_api::StarknetApiError),
+    State(blockifier::state::errors::StateError),
+    EntryPoint(blockifier::execution::errors::EntryPointExecutionError),
+    Transaction(blockifier::transaction::errors::TransactionExecutionError),
+    Custom(&'static str),
+}
```
pierre-l, line 11: I'm slightly concerned that always mapping by type may not help during debugging: it's good to know an error is an `std::io::Error` but it's crucial to know where it happens.
WIth such a mapping, I fear we may lack this info in the case of multiple places potentially throwing `std::io::Error`.

crates/cli/Cargo.toml
```diff
@@ -10,9 +10,14 @@ path = "src/main.rs"
 
 [dependencies]
 tokio = { workspace = true }
+beerus-api = { path = "../api", optional = true }
 beerus-core = { path = "../core" }
 beerus-rpc = { path = "../rpc" }
```
pierre-l, line 15: There must be a more expressive name than "api", especially when we already have a "rpc" crate.
If it takes care of tx execution, how about "execution"?

crates/api/src/rpc.rs
```diff
@@ -0,0 +1,380 @@
...
+                if req.id.is_some() {
+                    ret.push(res);
+                }
+            }
+            Ok(Json(Response::Batch(ret)))
+        }
+    }
+}
+
+#[allow(unused_variables)]
```
pierre-l, line 129: This should not be necessary.

crates/api/src/gen.rs
```diff
@@ -0,0 +1,9363 @@
...
+    pub struct Felt(String);
+
+    mod felt {
+        use super::jsonrpc;
+        use super::Felt;
+        use once_cell::sync::Lazy;
+        use regex::Regex;
+
+        static FELT_REGEX: Lazy<Regex> = Lazy::new(|| {
+            Regex::new("^0x(0|[a-fA-F1-9]{1}[a-fA-F0-9]{0,62})$").unwrap()
```
pierre-l, line 964: An expect would be better here.

crates/api/src/gen.rs
```diff
@@ -0,0 +1,9363 @@
...
+        impl AsRef<String> for U128 {
+            fn as_ref(&self) -> &String {
+                &self.0
+            }
+        }
+    }
+
+    #[derive(Clone, Debug, Deserialize, Serialize)]
+    #[serde(try_from = "String")]
+    pub struct U64(String);
```
pierre-l, line 1881: Why such newtypes?

crates/api/src/gen.rs
```diff
@@ -0,0 +1,9363 @@
...
+                    |args_by_pos| {
+                        let ArgByPos(block_id) = args_by_pos;
+                        ArgByName { block_id }
+                    },
+                )
+            });
+
+        let args: ArgByName = match args {
+            Ok(args) => args,
+            Err(_) => {
```
pierre-l, line 2907: Any chance we can log the cause here?

crates/api/src/gen.rs
```diff
@@ -0,0 +1,9363 @@
...
+                return jsonrpc::Response::error(-32602, "Invalid params")
+            }
+        };
+
+        let ArgByName { block_id } = args;
+
+        match rpc.getBlockWithTxs(block_id).await {
+            Ok(ret) => match serde_json::to_value(ret) {
+                Ok(ret) => jsonrpc::Response::result(ret),
+                Err(_) => jsonrpc::Response::error(-32603, "Internal error"),
```
pierre-l, line 2917: Any chance we can log the cause here?

crates/api/src/exe/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    let versioned_constants = VersionedConstants::latest_constants();
+
+    let block_context = BlockContext::new_unchecked(
+        &block_info,
+        &chain_info,
+        versioned_constants,
+    );
+
+    let tx_info = TransactionInfo::Deprecated(DeprecatedTransactionInfo {
```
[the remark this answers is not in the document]
sergey-melnychuk: Because the spec defines it.

.github/workflows/check.yml
```diff
@@ -24,6 +24,16 @@ jobs:
...
 
+  test-experimental:
+    runs-on: ubuntu-latest
+    env:
+      BEERUS_EXPERIMENTAL_TEST_STARKNET_URL: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo test -p beerus-api
```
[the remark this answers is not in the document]
sergey-melnychuk: Without the env var set, those tests don't actually run. This section runs the tests separately.

crates/api/src/exe/err.rs
sergey-melnychuk: Such mapping is way out of scope of this enum that only defines the error.
What you've described is a `Context`, which can be added as a layer on top with eyre/anyhow.

crates/api/src/gen.rs
sergey-melnychuk: All the generated code comes directly from the spec. The spec defines such type, so the generated code has it.

README.md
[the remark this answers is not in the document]
sergey-melnychuk: I consider this PR a single-topic one. The topic just appears to be huge.

crates/api/Cargo.toml
sergey-melnychuk: It is a raw experimental server (generated from the spec) that eventually will allow stateless transaction execution. The readme describes how to get the generated code from the supplied specs (which are slightly patched to become codegen-friendly).

crates/api/README.md
sergey-melnychuk: See comment above. Specific instructions how to generated it are provided.

crates/cli/Cargo.toml
sergey-melnychuk: Naming things is hard.

crates/api/src/gen.rs
sergey-melnychuk: Done

crates/api/src/gen.rs
sergey-melnychuk: Done

crates/api/src/gen.rs
sergey-melnychuk: Even though I don't thing `.expect()` makes a lot of difference compared to `.upwrap()`, I have also no problems with the change. Done.

crates/cli/Cargo.toml
sergey-melnychuk: Renamed to 'experimental-api`.

README.md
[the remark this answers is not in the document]
pierre-l: I see 3: 
- Cleanup in the README.md, .gitignore
- Stateless execution
- OpenRPC-based model generation

All of these should have been independent for a more effective review & process.

crates/api/src/exe/err.rs
pierre-l: It's not an `anyhow`/`eyre`-like context, the mapping to an `anyhow` error would happen subsequently, once this pieces of info are already lost.
It's fine for now, but I've used this pattern multiple times and systematically complained about it as soon as there are multiple sources of an IO error or transport error or anything.

crates/api/src/gen.rs
pierre-l: It's much better to know which regex failed to compile than just that the program panicked at the Nth line and to then have to dig in the Beerus code, especially generated code which tend to be less easy to digest.
This will also happen at runtime and will be user-facing in the context of Beerus. We shouldn't require our users to read our code.

crates/experimental-api/src/gen.rs
```diff
@@ -0,0 +1,9596 @@
+pub use gen::*;
+
+#[allow(clippy::module_inception)]
+// vvv GENERATED CODE BELOW vvv
+#[allow(dead_code)]
```
pierre-l, line 5: Making this file its own lib crate would avoid the need for this and _might_ help spot issues

crates/experimental-api/src/gen.rs
```diff
@@ -0,0 +1,9596 @@
...
+#[allow(non_snake_case)]
+#[allow(unused_variables)]
```
pierre-l, line 7: I'm not sure why this should be needed?

sergey-melnychuk: oops, got re-generated, will drop

sergey-melnychuk: done

crates/experimental-api/src/gen.rs
sergey-melnychuk: Since `clippy` actually doesn't complain with `allow(dead_code)` and `allow(unused_variables)` removed, I think we're good for now.

I prefer unit-testing, that *will* spot issues :)

crates/api/src/exe/mod.rs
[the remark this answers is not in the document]
sergey-melnychuk: No big reason behind choosing the `DeprecatedTransactionInfo` - just the example transaction I have used appears to be the "deprecated" (only by name, such transactions are still supported) one.

crates/api/Cargo.toml
sergey-melnychuk: I can't stress enough that this code is *experimental*. It is not replacing any 

> What values does iamgroot bring here if we already have a working RPC server?

I don't suggest replacing existing working RPC server. The value that iamgroot can *potentially* bring is the value of:

1. not wasting time for "spec fixes" anymore
1. seamless support of multiple spec versions
1. own DTO and network-related code (space for optimisations and customisations)
1. library vs. framework - code generated with iamgroot does not depend on any third-parties besides `serde_json`, which allows seamless migration between network frameworks.

> What value does stateless transaction execution bring?

Stateless (and trustless) querying & execution is literally the ultimate goal of this project. Once we can execute transactions in a stateless manner, we can actually start verifying intermediate storage values with merkle proofs - thus achieving "trustlessness" eventually. Without it the project makes no sense IMHO.

> How do you trust these without Helios?

(Please note I am not suggesting replacing Helios or doing anything with it. At this point this experimental code for transaction execution does not use Ethereum at all and thus does not need Helios. It likely will need Helios at some point, we're just not there yet).

Helios is an amazing tool, but it is not a silver bullet. There are alternative ways to ensure some level of "trust" (e.g. query multiple service providers, run own node, etc) that we did not explore nor consider. And I don't suggest to do it at this point. But again - at this point for this experimental code Helios is simply not necessary.

pierre-l (approved):

Issue #596, opened 2024-02-27
feat: RPC spec version check on startup


pierre-l:
Beerus should call `starknet_specVersion` on startup to make sure the backend Starknet node specified via the config uses the same spec.

A constant should be extracted from [this value](https://github.com/eigerco/beerus/blob/main/crates/rpc/src/api.rs#L528).

We should check if the Starknet OpenRPC spec effectively uses semver. If it's the case, we _might_ want to just enforce the same major version, or major+minor.

In the case of a mismatch, an error/fatal log should be printed and Beerus should exit.

psychiller:
I can take it up.
I will move spec_version value to config file from we read RPC endpoints. And then during `Client::new` function, a call can be made to starknet node and then check the result. If there is any better way, I could do it that way as well.

pierre-l:
> config file
I'd rather have this as a code constant since this version is dictated by the `starknet-rs` version in use in Beerus.

psychiller:
I was trying to check by implementing it and found we need to upgrade `starknet-rs` to new version. In `0.6.0` `spec_version` function is not available in provider.

pierre-l:
Ah, bummer.
The problem with more recent versions of `starknet-rs` is that they are based on [starknet-specs:v6.0.0](https://github.com/starkware-libs/starknet-specs/tree/v0.6.0), which AFAIK isn't supported by major API providers like Alchemy yet.
I expect support for the 0.6.0 spec will come anytime soon as 0.7 is already around the corner, but in the meantime our immediate position is to stay on 0.5.1 on the main branch.

pierre-l:
We could test more API providers from the [official list](https://docs.starknet.io/documentation/tools/api-services/), check which ones support 0.6.0 and if we have a few of these, migrate to 0.6.0 and list these providers in the README.md.

How does that sound @sergey-melnychuk?

psychiller:
Yeah, this should be straightforward now.

PR #603, opened 2024-02-29
feat: RPC spec version check on startup

psychiller:
Added check for spec version for beerus rpc client and remote starknet rpc client on rpc startup

### DIFF
--- a/crates/rpc/src/error.rs
+++ b/crates/rpc/src/error.rs
@@ -1,6 +1,7 @@
 use beerus_core::CoreError;
 use eyre::Report;
 use jsonrpsee::core::Error;
+use jsonrpsee::types::error::ErrorCode;
 use jsonrpsee::types::ErrorObjectOwned;
 use starknet::core::types::StarknetError;
 use starknet::providers::ProviderError;
@@ -11,8 +12,10 @@ use thiserror::Error;
 pub enum BeerusRpcError {
     #[error(transparent)]
     Provider(ProviderError),
-    #[error("{:?}", .0)]
+    #[error("{0:?}")]
     Other((i32, String)),
+    #[error("unexpected RPC spec version: {0}")]
+    UnexpectedSpecVersion(String),
 }
 
 impl From<BeerusRpcError> for ErrorObjectOwned {
@@ -195,6 +198,13 @@ impl From<BeerusRpcError> for ErrorObjectOwned {
             },
             BeerusRpcError::Other(other_err) => {
                 ErrorObjectOwned::owned(other_err.0, other_err.1, None::<()>)
+            }
+            BeerusRpcError::UnexpectedSpecVersion(err) => {
+                ErrorObjectOwned::owned(
+                    ErrorCode::InvalidParams.code(),
+                    err,
+                    None::<()>,
+                )
             }
         }
     }
--- a/crates/rpc/src/error.rs
+++ b/crates/rpc/src/error.rs
@@ -1,14 +1,18 @@
 use beerus_core::CoreError;
 use eyre::Report;
+use jsonrpsee::core::Error;
 use jsonrpsee::types::ErrorObjectOwned;
 use starknet::core::types::StarknetError;
 use starknet::providers::ProviderError;
 use starknet::providers::ProviderError::StarknetError as StarknetProviderError;
+use thiserror::Error;
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum BeerusRpcError {
+    #[error(transparent)]
     Provider(ProviderError),
-    Other((i32, String)),
+    #[error("Other: `{code}` :: `{msg}`")]
+    Other { code: i32, msg: String },
 }
 
 impl From<BeerusRpcError> for ErrorObjectOwned {
@@ -189,8 +193,8 @@ impl From<BeerusRpcError> for ErrorObjectOwned {
                     None::<()>,
                 ),
             },
-            BeerusRpcError::Other(other_err) => {
-                ErrorObjectOwned::owned(other_err.0, other_err.1, None::<()>)
+            BeerusRpcError::Other { code, msg } => {
+                ErrorObjectOwned::owned(code, msg, None::<()>)
             }
         }
     }
@@ -204,12 +208,18 @@ impl From<ProviderError> for BeerusRpcError {
 
 impl From<CoreError> for BeerusRpcError {
     fn from(err: CoreError) -> Self {
-        BeerusRpcError::Other((-32601, format!("{err}")))
+        BeerusRpcError::Other { code: -32601, msg: format!("{err}") }
     }
 }
 
 impl From<Report> for BeerusRpcError {
     fn from(err: Report) -> Self {
-        BeerusRpcError::Other((-32601, format!("{err}")))
+        BeerusRpcError::Other { code: -32601, msg: format!("{err}") }
+    }
+}
+
+impl From<Error> for BeerusRpcError {
+    fn from(err: Error) -> Self {
+        BeerusRpcError::Other { code: -32601, msg: format!("{err}") }
     }
 }
--- a/crates/rpc/src/error.rs
+++ b/crates/rpc/src/error.rs
@@ -11,8 +11,8 @@ use thiserror::Error;
 pub enum BeerusRpcError {
     #[error(transparent)]
     Provider(ProviderError),
-    #[error("Other: `{code}` :: `{msg}`")]
-    Other { code: i32, msg: String },
+    #[error("{:?}", .0)]
+    Other((i32, String)),
 }
 
 impl From<BeerusRpcError> for ErrorObjectOwned {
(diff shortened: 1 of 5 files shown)


crates/rpc/src/lib.rs
```diff
@@ -18,6 +21,27 @@ impl BeerusRpc {
     }
 
     pub async fn run(self) -> Result<(SocketAddr, ServerHandle), Error> {
+        let local_spec_version =
+            self.spec_version().await.map_err(BeerusRpcError::from).unwrap();
```
pierre-l, line 25: I'd be interested in getting rid of these `unwrap`s.
I suppose the one for the local version would just never happen, if that's the case then an `expect` saying it should never happen would be nicer.

crates/rpc/src/lib.rs
```diff
@@ -18,6 +21,27 @@ impl BeerusRpc {
...
+
+        let remote_spec_version = self
+            .beerus
+            .starknet_client
+            .spec_version()
+            .await
+            .map_err(BeerusRpcError::from)
+            .unwrap();
```
pierre-l, line 33: This should definitely be a log&return instead of an `unwrap`

etc/conf/beerus.json
```diff
@@ -1,5 +1,5 @@
 {
     "network": "MAINNET",
     "eth_execution_rpc": "https://eth-mainnet.g.alchemy.com/v2/<YOUR API KEY>",
-    "starknet_rpc": "https://starknet-mainnet.g.alchemy.com/v2/<YOUR API KEY>"
+    "starknet_rpc": " https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0.6/<YOUR API KEY>"
```
pierre-l, line 4: Thanks! Could you do the same in the TOML file?

pierre-l:
FTR, link to the original issue: #596

psychiller:
Added changes as requested. I used match instead of doing unwrap, if there is a better way, I would be happy to do that instead.

crates/rpc/src/lib.rs
```diff
@@ -18,6 +20,31 @@ impl BeerusRpc {
...
+                    error!(
+                        "Spec version mismatch between Beerus {} and remote Starknet RPC {}", local_spec_version, remote_version
+                    );
+                    return Err(Error::Custom(
+                        format!("Spec version mismatch between Beerus {} and remote Starkent RPC {}", local_spec_version, remote_version)
+                            .to_string(),
+                    ));
+                }
+            },
+            Err(_) => return Err(Error::Custom("Failed to fetch remote spec version".to_string())),
```
sergey-melnychuk, line 45: Why drop the cause of the error here, and not just fail fast when call to `spec_version` fails?

crates/rpc/src/lib.rs
```diff
@@ -18,6 +20,31 @@ impl BeerusRpc {
...
+            .beerus
+            .starknet_client
+            .spec_version()
+            .await;
+
+        // check remote spec version
+        match remote_spec_version {
+            Ok(remote_version) => {
+                if remote_version != local_spec_version {
+                    error!(
```
sergey-melnychuk, line 36: Duplicated error message (DRY). We can just return an error and let client code decide if it should be logged or not.

crates/rpc/src/lib.rs
```diff
@@ -18,6 +20,31 @@ impl BeerusRpc {
...
+        let local_spec_version =
+            self.spec_version().await.expect("Should never happen");
+
+        let remote_spec_version = self
+            .beerus
+            .starknet_client
+            .spec_version()
+            .await;
+
+        // check remote spec version
```
sergey-melnychuk, line 32: The need for this comment is a hint that a function can be extracted that checks that the spec versions match. And such function can be a subject for a unit-test coverage.

crates/rpc/src/lib.rs
```diff
@@ -18,6 +20,31 @@ impl BeerusRpc {
     }
 
     pub async fn run(self) -> Result<(SocketAddr, ServerHandle), Error> {
+        let local_spec_version =
+            self.spec_version().await.expect("Should never happen");
```
sergey-melnychuk, line 24: > Should never happen

If I had a nickel every time something that should never happen actually happens... I would have had a few nickels :) Why not add an error context and fail-fast (with `?`)?

One of the ways to proceed might be extracting the version as a constant and adding a unit-test for `spec_version()` to return a value that matches the constant. Then there is no need for a local call to `spec_version` and simple check against constant is enough.

crates/rpc/src/lib.rs
psychiller: `run` function is returning `jsonrpsee::Error` and error returned from call is `starknet::providers::ProviderError`
One way we can do it is implement another BeerusRPCError from trait in error.rs for `jsonrpsee::Error` and then use BeerusRPCError in return in `run` function.
If this sounds good, I can implement this or if there is another way to resolve it, I can do it that way.

crates/rpc/src/lib.rs
pierre-l: Hey, I'm the one who suggested that!

You're right though, a constant would avoid this.

sergey-melnychuk: I am not pushing for a constant, it is only "one of [many] ways to proceed".

It is `expect` that looks unusual. But I guess failing to return a static string should indeed never happen.

crates/rpc/src/lib.rs
sergey-melnychuk: Adding extra variant to `BeerusRpcError` for `jsonrpsee::Error` seems like a reasonable way to proceed. WDYT @pierre-l ?

psychiller: To make this approach work, we will need to implement `std::error::Error` which in turn required implementation of `std::fmt::Display` for BeerusRpcError. If we are going with this approach, any pointers on implementing these traits would be helpful

sergey-melnychuk: [thiserror](https://crates.io/crates/thiserror)

crates/rpc/src/error.rs
```diff
@@ -1,14 +1,18 @@
...
+use thiserror::Error;
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum BeerusRpcError {
+    #[error(transparent)]
     Provider(ProviderError),
-    Other((i32, String)),
+    #[error("Other: `{code}` :: `{msg}`")]
+    Other { code: i32, msg: String },
```
sergey-melnychuk, line 15: This is a breaking change to the public API. Is it necessary?

psychiller: To use thiserror, had to declare error message, and with current Other((i32, String)), I was not able to do that, because values are nested.
Alternative was to either separately implement Display trait for Other((i32, String)) or use a version like Other(i32, String) or what I implemented currently so values can be accessed

sergey-melnychuk: Spec mismatch seems to be specific case of an error, so it doesn't have to match `Other`. Particularly `code` and `message` (supposedly for JSON-RPC response) do not make sense when specs mismatch, as we can't even have a server running in such case. I'd add an enum variant to represent the spec mismatch situation, with respective error message derived by `#[error("...")]`. Then there is no need to change `Other` variant.

crates/rpc/src/error.rs
```diff
@@ -1,13 +1,17 @@
...
 use starknet::providers::ProviderError;
 use starknet::providers::ProviderError::StarknetError as StarknetProviderError;
+use thiserror::Error;
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum BeerusRpcError {
+    #[error(transparent)]
     Provider(ProviderError),
+    #[error("{:?}", .0)]
```
sergey-melnychuk, line 14:
```suggestion
    #[error("{0:?}")]
```

crates/rpc/src/error.rs
```diff
@@ -1,13 +1,17 @@
...
     Other((i32, String)),
 }
```
sergey-melnychuk, line 16:
```suggestion
    #[error("unexpected RPC spec version: {0}")]
    UnexpectedSpecVersion(String),
}
```

crates/rpc/src/error.rs
pierre-l: This looks go to me now. We could split the error type to have a specific one for the `run` method, avoiding the occurrence of this variant for regular RPC calls, but that can be done outside of this PR.

Thoughts @sergey-melnychuk ?

crates/rpc/src/error.rs
```diff
@@ -192,6 +199,13 @@ impl From<BeerusRpcError> for ErrorObjectOwned {
             BeerusRpcError::Other(other_err) => {
                 ErrorObjectOwned::owned(other_err.0, other_err.1, None::<()>)
             }
+            BeerusRpcError::UnexpectedSpecVersion(err) => {
```
sergey-melnychuk, line 202: This needs significant refactoring of errors project-wise, it would be way out of scope for this PR.

sergey-melnychuk (approved):
LGTM % (error organisation needs refactoring, but it is a topic for a different PR)

pierre-l (approved):

pierre-l:
Thanks @psychiller!

PR #606, opened 2024-03-04
fix: Make getEvents comply with the OpenRPC spec

pierre-l:
Our `starknet_getEvents` signature differs from the spec.

The spec has a single "filter" parameter including the actual filter parameters as well as the paging parameters.
Our implementation separates both.
This PR brings our implementation closer to the spec.

[Method spec](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_api_openrpc.json#L768C9-L814C11):
```json
{
            "name": "starknet_getEvents",
            "summary": "Returns all events matching the given filter",
            "description": "Returns all event objects matching the conditions in the provided filter",
            "params": [
                {
                    "name": "filter",
                    "summary": "The conditions used to filter the returned events",
                    "required": true,
                    "schema": {
                        "title": "Events request",
                        "allOf": [
                            {
                                "title": "Event filter",
                                "$ref": "#/components/schemas/EVENT_FILTER"
                            },
                            {
                                "title": "Result page request",
                                "$ref": "#/components/schemas/RESULT_PAGE_REQUEST"
                            }
                        ]
                    }
                }
            ],
            "result": {
                "name": "events",
                "description": "All the event objects matching the filter",
                "schema": {
                    "title": "Events chunk",
                    "$ref": "#/components/schemas/EVENTS_CHUNK"
                }
            },
            "errors": [
                {
                    "$ref": "#/components/errors/PAGE_SIZE_TOO_BIG"
                },
                {
                    "$ref": "#/components/errors/INVALID_CONTINUATION_TOKEN"
                },
                {
                    "$ref": "#/components/errors/BLOCK_NOT_FOUND"
                },
                {
                    "$ref": "#/components/errors/TOO_MANY_KEYS_IN_FILTER"
                }
            ]
        },
```

### DIFF
--- a/crates/rpc/src/api.rs
+++ b/crates/rpc/src/api.rs
@@ -9,8 +9,8 @@ use starknet::core::types::{
     BlockHashAndNumber, BlockId, BroadcastedDeclareTransaction,
     BroadcastedDeployAccountTransaction, BroadcastedInvokeTransaction,
     BroadcastedTransaction, ContractClass, DeclareTransactionResult,
-    DeployAccountTransactionResult, EventFilter, EventsPage, FeeEstimate,
-    FieldElement, FunctionCall, InvokeTransactionResult,
+    DeployAccountTransactionResult, EventFilterWithPage, EventsPage,
+    FeeEstimate, FieldElement, FunctionCall, InvokeTransactionResult,
     MaybePendingBlockWithTxHashes, MaybePendingBlockWithTxs,
     MaybePendingStateUpdate, MaybePendingTransactionReceipt, MsgFromL1,
     SimulationFlagForEstimateFee, SyncStatusType, Transaction,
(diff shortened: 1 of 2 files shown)


crates/rpc/src/api.rs
```diff
@@ -465,13 +463,15 @@ impl BeerusRpcServer for BeerusRpc {
 
     async fn get_events(
         &self,
-        filter: EventFilter,
-        continuation_token: Option<String>,
-        chunk_size: u64,
+        params: EventFilterWithPage,
```
sergey-melnychuk, line 466: The spec states the parameter name as `filter`. It will work for positional parameters, but not for named ones. This also looks like a good opportunity to provide missing test case :)

(A reference to the generated code: https://github.com/eigerco/beerus/blob/5f6860cb50cd98429716667da7d2abff1e85e35f/crates/api/src/gen.rs#L3039)

pierre-l: Correct. I saw that then forgot about it.

Now a problem we have is `starknet-rs` generates requests with positional parameters when the spec theoretically restricts these to named parameters only. But that's a problem for another day, and may become a non-issue if we switch to `iamgroot`.

sergey-melnychuk (approved):

PR #608, opened 2024-03-12
tests: stateless execution example

sergey-melnychuk:
Implement missing pieces for stateless call execution and add simple test case for it.

### DIFF
--- a/crates/experimental-api/src/exe/err.rs
+++ b/crates/experimental-api/src/exe/err.rs
@@ -4,6 +4,10 @@ use thiserror::Error as ThisError;
 pub enum Error {
     #[error("IO error: {0:?}")]
     Io(std::io::Error),
+    #[error("Base64 error: {0:?}")]
+    Base64(base64::DecodeError),
+    #[error("Serde error: {0:?}")]
+    Serde(serde_json::Error),
     #[error("Reqwest error: {0:?}")]
     Reqwest(reqwest::Error),
     #[error("Codegen error: {0:?}")]
@@ -25,6 +29,18 @@ pub enum Error {
 impl From<std::io::Error> for Error {
     fn from(error: std::io::Error) -> Self {
         Self::Io(error)
+    }
+}
+
+impl From<base64::DecodeError> for Error {
+    fn from(error: base64::DecodeError) -> Self {
+        Self::Base64(error)
+    }
+}
+
+impl From<serde_json::Error> for Error {
+    fn from(error: serde_json::Error) -> Self {
+        Self::Serde(error)
     }
 }
 
--- a/crates/experimental-api/src/exe/err.rs
+++ b/crates/experimental-api/src/exe/err.rs
@@ -3,83 +3,29 @@ use thiserror::Error as ThisError;
 #[derive(Debug, ThisError)]
 pub enum Error {
     #[error("IO error: {0:?}")]
-    Io(std::io::Error),
+    Io(#[from] std::io::Error),
     #[error("Base64 error: {0:?}")]
-    Base64(base64::DecodeError),
+    Base64(#[from] base64::DecodeError),
     #[error("Serde error: {0:?}")]
-    Serde(serde_json::Error),
+    Serde(#[from] serde_json::Error),
     #[error("Reqwest error: {0:?}")]
-    Reqwest(reqwest::Error),
+    Reqwest(#[from] reqwest::Error),
     #[error("Codegen error: {0:?}")]
     IamGroot(iamgroot::jsonrpc::Error),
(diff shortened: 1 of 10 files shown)


crates/experimental-api/tests/exe.rs
```diff
@@ -0,0 +1,41 @@
+use beerus_experimental_api::{
+    exe::exec,
+    gen::{client::blocking::Client, FunctionCall},
+};
+
+#[allow(dead_code)]
```
pierre-l, line 6: Is this necessary?

crates/experimental-api/tests/common/mod.rs
```diff
@@ -14,6 +17,24 @@ impl From<iamgroot::jsonrpc::Error> for Error {
...
+    fn from(error: beerus_experimental_api::exe::err::Error) -> Self {
+        Self::Exe(error)
+    }
+}
+
+impl From<serde_json::Error> for Error {
+    fn from(error: serde_json::Error) -> Self {
+        Self::Json(error)
+    }
+}
```
pierre-l, line 36: Since we have `thiserror` in this crate, all this code could be reduced to `#[from]` annotations.

crates/experimental-api/src/exe/mod.rs
```diff
@@ -126,33 +112,30 @@ pub fn exec(url: &str, txn: gen::BroadcastedTxn) -> Result<(), Error> {
...
 
-    let client = gen::client::blocking::Client::new(url);
     let diff = CommitmentStateDiff {
         storage_updates: Default::default(),
         address_to_nonce: Default::default(),
         address_to_class_hash: Default::default(),
         class_hash_to_compiled_class_hash: Default::default(),
     };
-    let mut proxy = StateProxy { client, diff };
+    let mut proxy = StateProxy { client: client.clone(), diff };
```
pierre-l, line 133: This clone could be skipped

crates/experimental-api/src/exe/err.rs
```diff
@@ -56,6 +74,14 @@ impl From<blockifier::transaction::errors::TransactionExecutionError>
...
+    fn from(
+        error: cairo_vm::types::errors::program_errors::ProgramError,
+    ) -> Self {
+        Self::Program(error)
+    }
+}
+
 impl From<iamgroot::jsonrpc::Error> for Error {
     fn from(error: iamgroot::jsonrpc::Error) -> Self {
         Self::IamGroot(error)
```
pierre-l, line 87: `#[from]` annotations should avoid the need for explicit `From` implementation

crates/experimental-api/src/exe/map.rs
```diff
@@ -29,7 +39,51 @@ impl TryFrom<StarkFelt> for gen::Felt {
...
 }
+
+fn build_contract_class(
+    class: DeprecatedContractClass,
+) -> Result<ContractClassV0, Error> {
+    let program = decode_program(class.program.as_ref())?;
+
+    let mut class = serde_json::to_value(class)?;
+    class["program"] = serde_json::from_str(&program)?;
+    let json = serde_json::to_string_pretty(&class)?;
```
pierre-l, line 64: We should probably use `serde_json::to_string` here to avoid unnecessary allocations.

Maybe something like this to keep the debug readability?
```
#[cfg(debug_assertions)]
let json = serde_json::to_string_pretty(&class)?;
#[cfg(not(debug_assertions))]
let json = serde_json::to_string(&class)?;
```

pierre-l (approved):

crates/experimental-api/tests/exe.rs
sergey-melnychuk: It just keeps slipping in :) Thank you for catching it!

sergey-melnychuk: Done.

crates/experimental-api/tests/common/mod.rs
sergey-melnychuk: Done.

crates/experimental-api/src/exe/mod.rs
sergey-melnychuk: Changed to `to_owned`.

crates/experimental-api/src/exe/err.rs
sergey-melnychuk: It actually forks for on all but `IamGroot` one :)

crates/experimental-api/src/exe/map.rs
sergey-melnychuk: `to_string` is more than enough here

PR #613, opened 2024-03-15
chore: Drop the `UnexpectedSpecVersion` error variant

pierre-l:
This error variant is only used on startup, it doesn't bring any value to have it exposed through the RPC interface.

### DIFF
--- a/crates/rpc/src/lib.rs
+++ b/crates/rpc/src/lib.rs
@@ -6,6 +6,7 @@ use std::net::SocketAddr;
 use api::{BeerusRpcServer, SPEC_VERION};
 use beerus_core::client::BeerusClient;
 use error::BeerusRpcError;
+use eyre::eyre;
 use jsonrpsee::server::{ServerBuilder, ServerHandle};
 use starknet::providers::Provider;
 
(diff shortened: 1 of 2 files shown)


crates/rpc/src/lib.rs
```diff
@@ -18,15 +19,14 @@ impl BeerusRpc {
         Self { beerus }
     }
 
-    pub async fn run(
-        self,
-    ) -> Result<(SocketAddr, ServerHandle), BeerusRpcError> {
+    pub async fn run(self) -> eyre::Result<(SocketAddr, ServerHandle)> {
```
sergey-melnychuk, line 22: We will have `eyre` tightly coupled to our public API. 
I don't know if this is a good idea.

crates/rpc/src/lib.rs
```diff
@@ -18,15 +19,14 @@ impl BeerusRpc {
...
         let remote_spec_version =
             self.beerus.starknet_client.spec_version().await?;
 
         if remote_spec_version != SPEC_VERION {
-            return Err(BeerusRpcError::UnexpectedSpecVersion(
-                remote_spec_version,
+            return Err(eyre!(
```
sergey-melnychuk, line 27: it would probably make message more informative if we provide what the expected version was

sergey-melnychuk:
IMHO spec version mismatch is a valid error variant and there is nothing wrong with it per se.

What *is* wrong is the unnecessary coupling between different error kinds: run and rpc ones.
I see the first step in organizing errors properly more like this: https://github.com/eigerco/beerus/pull/614

crates/rpc/src/lib.rs
pierre-l: Good point, we'll probably want to keep more informative errors.

PR #614, opened 2024-03-15
chore: decouple run and rpc errors

sergey-melnychuk:


crates/rpc/src/error.rs
```diff
@@ -18,17 +18,17 @@ pub enum RunError {
 }
 
 #[derive(Error, Debug)]
-pub enum BeerusRpcError {
+pub enum RpcError {
```
pierre-l: I'm all for it because I find this kind of prefix redundant, however this is a breaking change.

I wonder if we could add something like this?
```rust
#[deprecated]
use self::RpcError as BeerusRpcError;
```

crates/rpc/src/error.rs
```diff
@@ -1,27 +1,34 @@
...
 use jsonrpsee::types::ErrorObjectOwned;
 use starknet::core::types::StarknetError;
 use starknet::providers::ProviderError;
 use starknet::providers::ProviderError::StarknetError as StarknetProviderError;
 use thiserror::Error;
 
 #[derive(Error, Debug)]
-pub enum BeerusRpcError {
+pub enum RunError {
+    #[error("wrong RPC spec version: expected {1} but got {0}")]
```
pierre-l, line 12: We should clarify that the `got` version is the remote one.
Maybe we could also mention the relevant config key, as well as the "Version support" column from [this table](https://docs.starknet.io/documentation/tools/api-services/)?

PR #615, opened 2024-04-02
feat: use stateless execution in the rpc api

sergey-melnychuk:
This PR completes stateless execution implementation (with necessary dependency upgrades) and exposes it through the RPC API (existing unit-test coverage is satisfied).

### DIFF
--- a/crates/experimental-api/Cargo.toml
+++ b/crates/experimental-api/Cargo.toml
@@ -25,6 +25,7 @@ hex = "0.4.3"
 
 base64 = "0.22.0"
 flate2 = "1.0.28"
+cairo-lang-starknet-classes = "2.6.3"
 
 [dev-dependencies]
 tracing-subscriber.workspace = true
--- a/crates/experimental-api/Cargo.toml
+++ b/crates/experimental-api/Cargo.toml
@@ -2,7 +2,7 @@
 description = "JSON-RPC server & client auto-generated from the OpenRPC specification"
 edition = "2021"
 name = "beerus-experimental-api"
-version = "0.3.0"
+version = "0.1.0"
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
@@ -26,6 +26,7 @@ hex = "0.4.3"
 base64 = "0.22.0"
 flate2 = "1.0.28"
 cairo-lang-starknet-classes = "2.6.3"
+ureq = { version = "2.9.6", features = ["json"] }
 
 [dev-dependencies]
 tracing-subscriber.workspace = true
--- a/crates/experimental-api/Cargo.toml
+++ b/crates/experimental-api/Cargo.toml
@@ -18,10 +18,13 @@ tracing.workspace = true
 thiserror.workspace = true
 
 starknet-crypto = "0.6.1"
-blockifier = { git = "https://github.com/starkware-libs/blockifier", tag = "v0.5.0-rc.2" }
-starknet_api = "0.8.0"
+blockifier = { git = "https://github.com/starkware-libs/blockifier", tag = "v0.5.0" }
+starknet_api = "0.10.0"
 cairo-vm = "0.9.2"
 hex = "0.4.3"
 
 base64 = "0.22.0"
 flate2 = "1.0.28"
+
+[dev-dependencies]
+tracing-subscriber.workspace = true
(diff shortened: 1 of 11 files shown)


sergey-melnychuk:
Currently blocked by https://github.com/eigerco/beerus/issues/626

sergey-melnychuk:
The https://github.com/eigerco/beerus/pull/615/commits/69c88a7b7ce0a3c3d2c43d5ef9700d87230aa1ec brings proper support for both deprecated and regular (non-deprecated) contract classes calls, unblocks further integration of stateless execution into the RPC API.

crates/experimental-api/Cargo.toml
```diff
@@ -18,10 +18,15 @@ tracing.workspace = true
 thiserror.workspace = true
 
 starknet-crypto = "0.6.1"
-blockifier = { git = "https://github.com/starkware-libs/blockifier", tag = "v0.5.0-rc.2" }
-starknet_api = "0.8.0"
+blockifier = { git = "https://github.com/starkware-libs/blockifier", tag = "v0.5.0" }
```
pierre-l, line 21: Why not `blockifier = "0.5.0"`?

crates/experimental-api/Cargo.toml
```diff
@@ -18,10 +18,15 @@ tracing.workspace = true
...
+starknet_api = "0.10.0"
 cairo-vm = "0.9.2"
 hex = "0.4.3"
 
 base64 = "0.22.0"
 flate2 = "1.0.28"
+cairo-lang-starknet-classes = "2.6.3"
+ureq = { version = "2.9.6", features = ["json"] }
```
pierre-l, line 29: Why introduce a new HTTP client when the rest of the project relies on `reqwest`?

crates/experimental-api/tests/exe.rs
```diff
@@ -1,16 +1,17 @@
...
 mod common;
 
 #[test]
-fn test_exec() -> Result<(), common::Error> {
+fn test_call_deprecated_contract_class() -> Result<(), common::Error> {
     let Ok(url) = std::env::var("BEERUS_EXPERIMENTAL_TEST_STARKNET_URL") else {
         return Ok(());
     };
     let client = Client::new(&url);
+    tracing_subscriber::fmt::try_init().ok();
```
pierre-l, line 14: Are you sure you want to initial the subscriber here?
The `.ok()` feels unnecessary.
(I didn't know there was a `try_init` function, that's nice!)

crates/experimental-api/src/exe/err.rs
```diff
@@ -24,6 +25,8 @@ pub enum Error {
     ),
     #[error("Program error: {0:?}")]
     Program(#[from] cairo_vm::types::errors::program_errors::ProgramError),
+    #[error("Sierra compilation error: {0:?}")]
```
LKozlowski, line 28: This comment is unrelated to this PR, but what do you think about changing the error messages and making them start with lowercase instead of uppercase? In other parts of the code base, we also use error messages starting with lowercase, so that would make it consistent.

crates/experimental-api/src/rpc.rs
```diff
@@ -171,9 +173,27 @@ impl gen::Rpc for Context {
     async fn call(
         &self,
         request: FunctionCall,
-        block_id: BlockId,
+        _block_id: BlockId,
```
LKozlowski, line 176: nit: Do we still need the `block_id` parameter if we don't use it?

crates/experimental-api/Cargo.toml
sergey-melnychuk: https://github.com/eigerco/beerus/issues/626

crates/experimental-api/src/rpc.rs
sergey-melnychuk: We don't in this particular case, but it is a part of the RPC API

crates/experimental-api/src/exe/err.rs
sergey-melnychuk: https://github.com/eigerco/beerus/issues/636

sergey-melnychuk: https://github.com/eigerco/beerus/pull/637

crates/experimental-api/Cargo.toml
sergey-melnychuk: For some reason I was sure it wasn't available on crates.io. https://github.com/eigerco/beerus/pull/638

crates/experimental-api/tests/exe.rs
sergey-melnychuk: This allows running tests individually and get logs, while also allows running tests in bulk without errors of the subscriber being already initialized.

PR #612, opened 2024-03-15
Add `SEPOLIA` network configuration

sergey-melnychuk:
This PR adds Sepolia testnet configuration instead of Goerli one.

Respective Seploia support is necessary from helios part as well:
https://github.com/a16z/helios/pull/326

Issue #602, opened 2024-02-29
doc: Provide detailed instructions to check for `getProof` availability


pierre-l:
It is stated in the README.md that Beerus require both `eth_getProof` and `pathfinder_getProof`, but there are no instructions as to how to check their availability.

We should give CLI instructions for that.
We could also take this as another opportunity to check for Alchemy alternatives to recommend. See [this page](https://docs.starknet.io/documentation/tools/api-services/) for a list of candidates.

PR #617, opened 2024-04-04
Add RPC provider section to the README

LKozlowski:
This PR adds a RPC providers section to the README, which fixes [#602](https://github.com/eigerco/beerus/issues/602). I also reorganized it slightly, so I hope it is more readable now.

Also, since we introduced the `SEPOLIA` network here: https://github.com/eigerco/beerus/pull/612  there were still a few places that mentioned the deprecated `GORELI` network. I replaced all of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,20 +21,21 @@
 
 ## Getting Started
 
+### Running Beerus for the first time
+
+Copy the configuration file from `etc/conf/beerus.toml` and set up the RPC provider URLs in the copy.
+Make sure that providers are compatible. Read more about providers [here](#rpc-providers)
+
+Then run:
+```bash
+cargo run --release -- -c ./path/to/config.toml
+```
+
+Once Beerus has started to verify that everything is working correctly, run this command:
+```
+hurl etc/rpc/starknet_getStateRoot.hurl
+```
 ### Configuration
-
-Beerus relies on TWO untrusted RPC endpoints, one for L1 (Ethereum), and one for L2 (Starknet). 
-As these are untrusted they will typically not be nodes run on your local host or your local network.
-
-Beerus requires the [v0.6.0 of the Starknet OpenRPC specs](https://github.com/starkware-libs/starknet-specs/tree/v0.6.0).
-
-These untrusted RPC providers must adhere to both the L1 `eth_getProof` endpoint
-as well as the L2 `pathfinder_getProof` endpoint ([detailed instructions needed!](https://github.com/eigerco/beerus/issues/602)).For this we recommend using
-[Alchemy](https://docs.alchemy.com/reference/starknet-api-faq#what-versions-of-starknet-api-are-supported) as your untrusted L2 node provider. 
-
-More API providers can be found [here](https://docs.starknet.io/documentation/tools/api-services/).
-
-*NOTE: we rely on [helios](https://github.com/a16z/helios) for both valid checkpoint values and consensus rpc urls*
 
 | field   | example | description |
 | ----------- | ----------- | ----------- |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -21,42 +21,139 @@
...
+
+Etherum RPC provider must support the `eth_getProof` endpoint.
+
+You can check if the provider is compatible by running this command:
 ```bash
-cargo run --release -- -c ./path/to/config.toml
+# This is an example RPC url. Use your RPC provider url to check if the node is compatible.
+ETHERUM_RPC_URL="https://eth-sepolia.g.alchemy.com/v2/{YOUR_API_KEY}"
+curl --request POST \
+     --url $ETHERUM_RPC_URL \
```
pierre-l, line 76: Typos: `ETHERUM` -> +E

README.md
```diff
@@ -21,42 +21,139 @@
...
+```
 
-Then run:
+#### RPC providers
+Beerus relies on TWO untrusted RPC endpoints, one for L1 (Ethereum), and one for L2 (Starknet).
+As these are untrusted they will typically not be nodes run on your local host or your local network.
+
+##### Etherum RPC endpoint
+
+Etherum RPC provider must support the `eth_getProof` endpoint.
```
pierre-l, line 69: Typos: `Etherum` -> `Ethereum`

README.md
```diff
@@ -21,42 +21,139 @@
...
+# This is an example RPC url. Use your RPC provider url to check if the node is compatible.
+ETHERUM_RPC_URL="https://eth-sepolia.g.alchemy.com/v2/{YOUR_API_KEY}"
+curl --request POST \
+     --url $ETHERUM_RPC_URL \
+     --header 'content-type: application/json' \
+     --data '
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "eth_getProof",
```
sergey-melnychuk, line 82: I don't think we actually use `eth_getProof` anywhere directly. It is also a part of the Ethereum RPC API, so it should be supported by all service providers.

README.md
```diff
@@ -21,42 +21,139 @@
...
+hurl etc/rpc/starknet_getStateRoot.hurl
+```
+### Configuration
 
 | field   | example | description |
 | ----------- | ----------- | ----------- |
-| network | MAINNET or GOERLI | network to query |
-| eth_execution_rpc | https://eth-mainnet.g.alchemy.com/v2/YOURAPIKEY | untrusted l1 node provider url |
-| starknet_rpc | https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0.6/YOURAPIKEY | untrusted l2 node provider url |
+| network | MAINNET or SEPOLIA| network to query |
```
sergey-melnychuk, line 42: good catch 👍🏻

README.md
LKozlowski: Thanks! Fixed all typos :)

README.md
pierre-l: I'm pretty sure you're correct on both points, my bad.

LKozlowski: I'll remove that section then and just write a few words that we're using standard ethereum rpc api so all service providers will work.

LKozlowski: Just for reference here is a [link](https://github.com/a16z/helios?tab=readme-ov-file#supported-ethereum-execution-api-providers-) to helios readme about that endpoint. Maybe it used to be an issue, but is not anymore?

sergey-melnychuk: As long as we're good with any Ethereum provider that Helios is good with, I believe the link above is enough.

sergey-melnychuk (approved):

PR #624, opened 2024-04-12
test: starknet_blockHashAndNumber integration test

pierre-l:
Still as a draft as I saw this fail. It failed once in dozens of runs, but still, it shouldn't have.


pierre-l:
Confirmation it's not just a fluke: https://github.com/pierre-l/beerus-ci-test/actions/runs/8868946027/job/24349084059

I fear this inconsistency here might come directly from Pathfinder.

PR #628, opened 2024-04-12
tests(rpc): clean up fixtures for integration tests

sergey-melnychuk:
This PR removes dysfunctional caching in the integration test suite and changes the test fixtures to a stream-based (potentially more async-friendly) implementation.

Most comfortable way to review is commit-by-commit.

### DIFF
--- a/crates/rpc/tests/rpc.rs
+++ b/crates/rpc/tests/rpc.rs
@@ -102,6 +102,7 @@ mod blocks {
 mod fixtures {
     use super::*;
 
+    // TODO(#643): consider adding a cache here if necessary
     pub async fn latest_block(
     ) -> (JsonRpcClient<HttpTransport>, BlockWithTxs, BlockId) {
         let block = blocks::head().await.expect("failed to pull latest block");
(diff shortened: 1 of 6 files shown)


sergey-melnychuk:
Possible follow up: https://github.com/eigerco/beerus/issues/643

crates/rpc/tests/rpc.rs
```diff
@@ -1,192 +1,147 @@
...
-async fn latest_block_context() -> TestContext<()> {
-    context(|_block| Some(())).await
-}
-
-/// Creates a `TestContext` with the latest block with at least N transactions.
-async fn n_txs_context(min_tx_number: usize) -> TestContext<()> {
-    context(|block| {
-        if block.transactions.len() >= min_tx_number {
-            Some(())
+    async fn next(state: State) -> Option<(BlockWithTxs, State)> {
```
LKozlowski, line 36:
```suggestion
    async fn next_state(state: State) -> Option<(BlockWithTxs, State)> {
```
nit: I was confused at first and thought that the function produces the next block because it is in the module `blocks::next`, but then I realized that it is supposed to produce the next state for the unfold, hence my name change suggestion 😄

LKozlowski (approved):
Looks good! I like the approach with the streams. It looks clean.

crates/rpc/tests/rpc.rs
sergey-melnychuk: `change_stream_state` it is now

PR #630, opened 2024-04-16
ci: Run the ITs against multiple API providers

pierre-l:
Run the ITs against Alchemy, Reddio and Chainstack.

Keep Alchemy as the default check for a git push.
Have the other two run periodically.

Bound the health check loop to a max timeout of 60s.
Increase the job timeout to 30 minutes. 15 minutes leaves too little room for the unexpected.

Although these all rely on Pathfinder, this may still help determine if any of these is more reliable than the others, and will offer us redundancy in the case of an incident on Alchemy's side. It's good for CI, but also for Beerus users.

Secrets & variables will have to be updated before merging this PR.

### DIFF
--- a/.github/workflows/rust-integration-tests-alchemy.yml
+++ b/.github/workflows/rust-integration-tests-alchemy.yml
@@ -1,4 +1,4 @@
-name: rust-integration-tests
+name: rust-integration-tests-alchemy
 on:
   schedule:
     - cron: "0 8,20 * * *"
@@ -8,7 +8,7 @@ on:
 jobs:
   integration:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
@@ -19,11 +19,10 @@ jobs:
 
       - name: start beerus RPC on MAINNET
         env:
-          BEERUS_VERSION: "ci"
-          NETWORK: "mainnet"
           ETH_EXECUTION_RPC: ${{ secrets.ETH_EXECUTION_RPC_MAINNET }}
-          STARKNET_RPC: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET }}
-        run: ./target/release/beerus & while ! timeout 1 bash -c "echo > /dev/tcp/localhost/3030" 2> /dev/null; do sleep 2; done
+          STARKNET_RPC: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET_ALCHEMY }}
+        # Start Beerus, then try to reach the port every second for 60 seconds.
+        run: ./target/release/beerus & for i in {1..60}; do timeout 1 bash -c "echo > /dev/tcp/localhost/3030" 2> /dev/null && break || sleep 1; done
 
       - name: run integration test on MAINNET
         run: cargo test --package beerus-rpc --features integration-tests
--- a/.github/workflows/rust-integration-tests-chainstack.yml
+++ b/.github/workflows/rust-integration-tests-chainstack.yml
@@ -1,0 +1,31 @@
+name: rust-integration-tests-chainstack
+on:
+  schedule:
+    - cron: "0 8,20 * * *"
+  workflow_dispatch:
+
+jobs:
+  integration:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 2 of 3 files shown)


.github/workflows/rust-integration-tests-alchemy.yml
```diff
@@ -22,12 +22,15 @@ jobs:
           BEERUS_VERSION: "ci"
           NETWORK: "mainnet"
           ETH_EXECUTION_RPC: ${{ secrets.ETH_EXECUTION_RPC_MAINNET }}
-          STARKNET_RPC: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET }}
-        run: ./target/release/beerus & while ! timeout 1 bash -c "echo > /dev/tcp/localhost/3030" 2> /dev/null; do sleep 2; done
+          STARKNET_RPC: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET_ALCHEMY }}
```
pierre-l, line 25: If approved, set up the new secret before merging.

.github/workflows/rust-integration-tests-chainstack.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: build beerus
+        run: cargo build -p beerus-cli --release
+
+      - name: start beerus RPC on MAINNET
+        env:
+          BEERUS_VERSION: "ci"
+          NETWORK: "mainnet"
+          ETH_EXECUTION_RPC: ${{ secrets.ETH_EXECUTION_RPC_MAINNET }}
+          STARKNET_RPC: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET_CHAINSTACK }}
```
pierre-l, line 24: If approved, set up the new secret before merging.

.github/workflows/rust-integration-tests-reddio.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: build beerus
+        run: cargo build -p beerus-cli --release
+
+      - name: start beerus RPC on MAINNET
+        env:
+          BEERUS_VERSION: "ci"
+          NETWORK: "mainnet"
+          ETH_EXECUTION_RPC: ${{ secrets.ETH_EXECUTION_RPC_MAINNET }}
+          STARKNET_RPC: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET_REDDIO }}
```
pierre-l, line 24: If approved, set up the new secret before merging.

crates/rpc/tests/rpc.rs
```diff
@@ -23,11 +23,13 @@ lazy_static::lazy_static! {
     static ref BLOCK_CACHE: Mutex<SizedCache<u64, MaybePendingBlockWithTxs>> = Mutex::new(SizedCache::with_size(1000));
 }
 
+fn rpc_url() -> Url {
+    let config = Config::from_env();
```
sergey-melnychuk, line 27: (follow up on https://github.com/eigerco/beerus/pull/631 && https://github.com/eigerco/beerus/pull/631#issuecomment-2066467473)

> I really wish is was possible to fix one little thing without having to fix everything that's around, significantly extending the scope of an initially trivial PR, and bringing multiple breaking changes in the process. I'm closing this PR.
> 
> #630 will use config files.

---

_(Just trying to understand the bigger picture behind this PR and #630 by asking some questions. They are not suggestions, just questions, no need to act on them. But answers to those questions I believe might be useful to choose the rational way to proceed here.)_

Why require to parse the full config here if only RPC URL is necessary? Won't it be easier to just lookup necessary env var for the RPC URL or am I missing something?

With env var lookup even the extraction of the `rpc_url` function is not necessary at this point, as it is only being used once (unless I am missing something here too).

With such env var simplification I fail to see the necessity of https://github.com/eigerco/beerus/pull/631 as a prerequisite for this PR. Please enlighten me if I am indeed missing the bigger picture here.

pierre-l: This env var doesn't exist, that's why I initially looked into #631.

Now these workflows could run sequentially, but it makes the overall assessment more tedious: I can only have them run about 12 times in a day and the reliability assessment has less value.
If I run them in parallel, I would either need 3 self-hosted runners, or different ports, or would quickly run out of GitHub actions credits.

pierre-l: Hm, it looks like I would need 3 runners anyway, making the port irrelevant.

pierre-l: I removed these changes as they are now obsolete.

sergey-melnychuk (approved):

PR #631, opened 2024-04-17
feat: Extend support for env-based configuration

pierre-l:
BREAKING CHANGE: `Config::from_env` now returns an `eyre::Result` instead of panicking on site.

The primary motivation is #630 , but I also wanted more explicit messages when running Beerus without specifying the `STARKNET_RPC` and `ETH_EXECUTION_RPC` keys.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -9,9 +9,9 @@ pub struct Args {
 }
 
 pub fn get_config(args: Args) -> eyre::Result<Config> {
-    Ok(if let Some(path) = args.config.as_ref() {
-        Config::from_file(path)?
+    if let Some(path) = args.config.as_ref() {
+        Config::from_file(path)
     } else {
         Config::from_env()
-    })
+    }
 }
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -85,35 +85,34 @@ impl Default for Config {
 
 impl Config {
     pub fn from_env() -> Result<Self> {
-        fn require_env_var(var_key: &'static str) -> Result<String> {
-            std::env::var(var_key).map_err(|_| eyre!("The \"{}\" env var must be set or a configuration file must be specified", var_key))
-        }
+        Self::from_vars(|key| std::env::var(key).ok())
+    }
+
+    fn from_vars<F>(get: F) -> Result<Self>
+    where
+        F: Fn(&'static str) -> Option<String> + 'static,
+    {
+        let require = |var_key: &'static str| {
+            get(var_key).ok_or_else(|| eyre!("The \"{}\" env var must be set or a configuration file must be specified", var_key))
+        };
 
         Ok(Self {
-            network: Network::from_str(
-                &std::env::var("NETWORK").unwrap_or_default(),
-            )
-            .unwrap_or(Network::MAINNET),
-            eth_execution_rpc: require_env_var("ETH_EXECUTION_RPC")?,
-            starknet_rpc: require_env_var("STARKNET_RPC")?,
-            data_dir: PathBuf::from(
-                std::env::var("DATA_DIR").unwrap_or_default(),
-            ),
-            poll_secs: u64::from_str(
-                &std::env::var("POLL_SECS").unwrap_or_default(),
-            )
-            .unwrap_or(DEFAULT_POLL_SECS),
-            rpc_addr: if let Ok(addr) = std::env::var("RPC_ADDR") {
-                SocketAddr::from_str(&addr)
-                    .context("Invalid value for `RPC_ADDR`")?
-            } else {
-                rpc_addr()
+            network: Network::from_str(&get("NETWORK").unwrap_or_default())
+                .unwrap_or(Network::MAINNET),
+            eth_execution_rpc: require("ETH_EXECUTION_RPC")?,
+            starknet_rpc: require("STARKNET_RPC")?,
+            data_dir: PathBuf::from(get("DATA_DIR").unwrap_or_default()),
+            poll_secs: u64::from_str(&get("POLL_SECS").unwrap_or_default())
+                .unwrap_or(DEFAULT_POLL_SECS),
+            rpc_addr: match get("RPC_ADDR") {
+                Some(addr) => SocketAddr::from_str(&addr)
+                    .context("Invalid value for `RPC_ADDR`")?,
+                None => rpc_addr(),
             },
-            fee_token_addr: if let Ok(addr) = std::env::var("FEE_TOKEN_ADDR") {
-                FieldElement::from_hex_be(&addr)
-                    .context("Invalid value for `FEE_TOKEN_ADDR`")?
-            } else {
-                fee_token_addr()
+            fee_token_addr: match get("FEE_TOKEN_ADDR") {
+                Some(addr) => FieldElement::from_hex_be(&addr)
+                    .context("Invalid value for `FEE_TOKEN_ADDR`")?,
+                None => fee_token_addr(),
             },
         })
     }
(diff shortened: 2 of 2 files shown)


crates/core/src/config.rs
```diff
@@ -84,25 +84,37 @@ impl Default for Config {
...
             poll_secs: u64::from_str(
                 &std::env::var("POLL_SECS").unwrap_or_default(),
             )
             .unwrap_or(DEFAULT_POLL_SECS),
-            rpc_addr: rpc_addr(),
-            fee_token_addr: fee_token_addr(),
-        }
+            rpc_addr: std::env::var("RPC_ADDR")
+                .map(|addr| {
+                    SocketAddr::from_str(&addr).expect("a valid socket address")
```
pierre-l, line 108: Unnecessary `expect`

sergey-melnychuk: `map_err` here?

crates/core/src/config.rs
```diff
@@ -84,25 +84,37 @@ impl Default for Config {
...
+
+        Ok(Self {
             network: Network::from_str(
                 &std::env::var("NETWORK").unwrap_or_default(),
             )
             .unwrap_or(Network::MAINNET),
-            eth_execution_rpc: std::env::var("ETH_EXECUTION_RPC")
-                .unwrap_or_default(),
-            starknet_rpc: std::env::var("STARKNET_RPC").unwrap_or_default(),
+            eth_execution_rpc: require_env_var("ETH_EXECUTION_RPC")?,
```
sergey-melnychuk, line 97: Extra sanity checks for RPC URLs might be useful here, just to fail fast if invalid config param was provided

sergey-melnychuk (commented):
Some unit-tests would be a nice addition to solidify configuration invariants

crates/core/src/config.rs
pierre-l: Yes, we probably could also change the type of the *_RPC vars to proper `Url`s instead of `String`s.

What do you think about me handling this in a separate issue?
Also I'm bothered with the *_RPC suffix, what do you think about changing this to *_URL?

crates/core/src/config.rs
pierre-l: I went with a dumb `if else`

[related push](https://github.com/eigerco/beerus/compare/35d374ee3f9928972ba9c1b0d68b2b5782dbccb6..c030b7d910e5da24ff7462347e60dd9f37fd6544)

crates/core/src/config.rs
LKozlowski: I think that you shouldn't remove the `_rpc` suffix as it tells what it really is, but at the same time I think that adding `_url` to it would be beneficial too, so in the end you would end up in something like `eth_execution_rpc_url`.

crates/core/src/config.rs
```diff
@@ -84,25 +84,38 @@ impl Default for Config {
...
+            starknet_rpc: require_env_var("STARKNET_RPC")?,
             data_dir: PathBuf::from(
                 std::env::var("DATA_DIR").unwrap_or_default(),
```
sergey-melnychuk, line 100: Maybe for separate PR: sanity check that directory exists and is writable seems reasonable here

crates/core/src/config.rs
```diff
@@ -84,25 +84,38 @@ impl Default for Config {
...
             ),
             poll_secs: u64::from_str(
                 &std::env::var("POLL_SECS").unwrap_or_default(),
```
sergey-melnychuk, line 103: Maybe separate PR: provide reasonable range (e.g. min as default, max for sanity check) here

crates/core/src/config.rs
sergey-melnychuk: Both params are passed as strings later on if I'm not mistaken, so I think some basic URL validation seems reasonable here.

I personally don't care much about naming and suffixes, but any changes there require respective update of readme and/or other docs to avoid users wondering.

crates/core/src/config.rs
sergey-melnychuk: https://github.com/eigerco/beerus/issues/632

crates/core/src/config.rs
sergey-melnychuk: https://github.com/eigerco/beerus/issues/632

sergey-melnychuk (commented):
Just to be clear: this PR looks acceptable and I assume works right but needs unit-tests coverage for configuration factory code. To achieve that decoupling from `std::env::var` seems to be necessary.

pierre-l:
@sergey-melnychuk @LKozlowski I added a [second commit](https://github.com/eigerco/beerus/pull/631/commits/228df0120a3c1e1311d90018b3463a55942a187d) for the unit tests
(also thanks for the issue & resolved threads)

crates/core/src/config.rs
```diff
@@ -84,25 +84,37 @@ impl Default for Config {
...
+    fn from_vars<F>(get: F) -> Result<Self>
+    where
+        F: Fn(&'static str) -> Option<String> + 'static,
+    {
+        let require = |var_key: &'static str| {
+            get(var_key).ok_or_else(|| eyre!("The \"{}\" env var must be set or a configuration file must be specified", var_key))
+        };
+
+        Ok(Self {
+            network: Network::from_str(&get("NETWORK").unwrap_or_default())
```
sergey-melnychuk, line 100: if the provided network is invalid, then mainnet is used as a default which might not be always desired behavior; I think an error must be returned when provided network is invalid - with respective test case.

crates/cli/src/lib.rs
```diff
@@ -9,9 +9,9 @@ pub struct Args {
 }
 
 pub fn get_config(args: Args) -> eyre::Result<Config> {
-    Ok(if let Some(path) = args.config.as_ref() {
-        Config::from_file(path)?
+    if let Some(path) = args.config.as_ref() {
+        Config::from_file(path)
     } else {
         Config::from_env()
```
sergey-melnychuk, line 15:
```suggestion
    args.config
        .map(|path| Config::from_file(&path))
        .or_else(Config::from_env)
```

sergey-melnychuk: Nevermind

crates/core/src/config.rs
```diff
@@ -84,25 +84,37 @@ impl Default for Config {
...
+                .unwrap_or(Network::MAINNET),
+            eth_execution_rpc: require("ETH_EXECUTION_RPC")?,
+            starknet_rpc: require("STARKNET_RPC")?,
+            data_dir: PathBuf::from(get("DATA_DIR").unwrap_or_default()),
+            poll_secs: u64::from_str(&get("POLL_SECS").unwrap_or_default())
```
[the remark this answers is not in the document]
pierre-l: This is the original behavior. It's out of scope of this PR. It doesn't help with with my original task. It would be another breaking change.

pierre-l:
I really wish is was possible to fix one little thing without having to fix everything that's around, significantly extending the scope of an initially trivial PR, and bringing multiple breaking changes in the process.
I'm closing this PR.

#630 will use config files.

sergey-melnychuk:
> I really wish is was possible to fix one little thing without having to fix everything that's around, significantly extending the scope of an initially trivial PR, and bringing multiple breaking changes in the process. I'm closing this PR.

I guess I am sometimes getting carried away with "ownership mentality" over the project and [the boy scout rule](https://deviq.com/principles/boy-scout-rule) ("Leave your code better than you found it.")

Yet I personally believe there is still value to improve the code at every opportunity provided. If the original behavior allows invalid configuration parameters to be propagated through the system, I don't see why we should stick to it and not improve it in place (with just a few lines of code btw). If some test cases are missing for the code being changed, why not fill this gap for the team and the project to benefit?

No one forces us to stick to the "original behavior". No one forces us not to to extend the scope of the PR slightly to improve code quality and test coverage. But what I believe what we as a team force each other to do (and keep each other accountable for doing it) is to stick to the highest code quality standards possible every time, every LoC and every PR.

> #630 will use config files.

If I'm being completely honest, I don't see why it might be a problem. https://github.com/eigerco/beerus/pull/630/files#r1574811607

Issue #639, opened 2024-04-26
bug: `Event` DTO is missing `from_address` field


sergey-melnychuk:
https://github.com/eigerco/beerus/blob/c51b34284a585e70666a699d55b09986e36bc8e2/crates/experimental-api/src/gen.rs#L861

https://github.com/eigerco/beerus/blob/c51b34284a585e70666a699d55b09986e36bc8e2/etc/spec/starknet/0.6.0/starknet_query_api_openrpc.json#L943

cc @pierre-l

PR #640, opened 2024-04-26
fix(spec): avoid schema entries shadowing

sergey-melnychuk:
Closes https://github.com/eigerco/beerus/issues/639

### DIFF
--- a/crates/experimental-api/src/gen.rs
+++ b/crates/experimental-api/src/gen.rs
@@ -858,7 +858,11 @@ pub mod gen {
     }
 
     #[derive(Clone, Debug, Deserialize, Serialize)]
-    pub struct Event(pub EventContent);
+    pub struct Event {
+        pub from_address: Address,
+        #[serde(flatten)]
+        pub event_content: EventContent,
+    }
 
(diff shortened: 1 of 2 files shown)


crates/experimental-api/src/gen.rs
```diff
@@ -1305,7 +1309,7 @@ pub mod gen {
         #[serde(default)]
         pub order: Option<i64>,
         #[serde(flatten)]
-        pub event: Event,
```
sergey-melnychuk, line 1308: this was another problem caused by the shadowing `EVENT` with `EVENT_CONTENT`...

sergey-melnychuk:
The `iamgroot` did work fine for the isolated definition if the `EVENT`: https://github.com/sergey-melnychuk/iamgroot/commit/48d88ad536e522e4b139783577c028a2384ac12e

(Symmetric fix in `iamgroot`: https://github.com/sergey-melnychuk/iamgroot/commit/97e8689c9d901a1c9497011e96ca90065554f847)

pierre-l (approved):
We'll probably have to keep that in mind when migrating to 0.7

LKozlowski (approved):

PR #641, opened 2024-04-29
fix: Expose the experimental server to the same URL as Beerus

pierre-l:
The experimental server expose its JSON-RPC interface at the path `/rpc` while Beerus does that at the root path.

This PR makes the experimental server use the root path as well so both behaviors are consistent.


sergey-melnychuk:
I'm not sure what is the purpose of this change.
I was imagining that experimental server might serve the page that exposes WASM-based Beerus implementation on the `/` endpoint, then maybe some user-friendly documentation in `/docs` and in such setup serving RPC from separate endpoint (e.g. `/rpc`) makes sense to me.

PR #645, opened 2024-05-17
feat(config): add validation

ICavlek:
Workflow summary

1. Introduced Validator crate that has trait Validate which can be used on Structs and it's attributes
2. Check for poll_secs
3. Check for URLs - Check whether the String is ok and perform a post on the url. If the returned http answer is ok, than the url is valid
4. Check for data dir - If the folder exists, check if is read only or not. If it doesn't exist, trace a warning.
5. Unit tests for behaviour validation

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -1,6 +1,7 @@
+use std::ffi::OsStr;
 use std::fs;
 use std::net::SocketAddr;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use ethers::types::Address;
@@ -16,6 +17,7 @@ use serde::Deserialize;
 use starknet::core::types::FieldElement;
 use starknet::providers::jsonrpc::{HttpTransport, JsonRpcClient};
 use url::Url;
+use validator::Validate;
 
 // defaults
 pub const DEFAULT_DATA_DIR: &str = "tmp";
@@ -36,15 +38,23 @@ pub const SEPOLIA_CONSENSUS_RPC: &str =
     "http://unstable.sepolia.beacon-api.nimbus.team";
 pub const SEPOLIA_FALLBACK_RPC: &str = "https://sync-sepolia.beaconcha.in";
 
+enum Blockchain {
+    Ethereum,
+    Starknet,
+}
+
 /// global config
-#[derive(Clone, Deserialize, Debug)]
+#[derive(Clone, Deserialize, Debug, Validate)]
 pub struct Config {
     pub network: Network,
+    #[validate(url)]
     pub eth_execution_rpc: String,
+    #[validate(url)]
     pub starknet_rpc: String,
     #[serde(default = "data_dir")]
     pub data_dir: PathBuf,
     #[serde(default = "poll_secs")]
+    #[validate(range(min = 1, max = 10))]
     pub poll_secs: u64,
     #[serde(default = "rpc_addr")]
     pub rpc_addr: SocketAddr,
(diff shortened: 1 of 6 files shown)


crates/core/src/config.rs
```diff
@@ -107,15 +117,72 @@ impl Config {
...
-        } else {
-            Err(eyre!("Unsupported config file format"))
+    pub async fn validate_params(&self) -> Result<()> {
+        self.validate()?;
+        Self::check_url(&self.eth_execution_rpc, Blockchain::Ethereum).await?;
+        Self::check_url(&self.starknet_rpc, Blockchain::Starknet).await?;
+        self.validate_data_dir()
+    }
+
+    async fn check_url(url: &str, blockchain: Blockchain) -> Result<()> {
```
LKozlowski, line 149: I'm not sure if sending HTTP requests is the right approach when we construct config objects. I think that we should only validate URLs in terms that they conform to the URLs spec. That way we check if the provided values are somewhat valid without introducing side effects and latency by sending a request, and beerus will fail to start anyway if we provide the wrong urls. What do you think?

crates/core/src/config.rs
```diff
@@ -36,15 +38,23 @@ pub const SEPOLIA_CONSENSUS_RPC: &str =
...
 pub struct Config {
     pub network: Network,
+    #[validate(url)]
     pub eth_execution_rpc: String,
+    #[validate(url)]
     pub starknet_rpc: String,
     #[serde(default = "data_dir")]
     pub data_dir: PathBuf,
     #[serde(default = "poll_secs")]
+    #[validate(range(min = 1, max = 10))]
```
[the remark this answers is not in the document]
ICavlek: I have fixed it to a range of 10 as a starting point for discussion. I am also not sure which value it should be, please add suggestion

crates/core/src/config.rs
ICavlek: I wanted to have it as an additional layer check. If it is unnecessary, I can easily remove it. @sergey-melnychuk, please add your opinion

crates/core/src/config.rs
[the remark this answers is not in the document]
sergey-melnychuk: The goal is to avoid unreasonably high poll rate that would still be a valid `u64` value. Maybe 1h (3600 seconds) is such a reasonable bound. We can always update it if necessary.

crates/core/src/config.rs
sergey-melnychuk: Functional check is a good idea I think, but it can be done after the valid config is created.

It is actually a very good check - make sure that both Ethereum and Starknet endpoint use the same network. We might want to call `*_chainId` on both and see if the results are compatible.

crates/core/src/config.rs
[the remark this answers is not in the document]
ICavlek: Ok, will set min = 1, max = 3600

sergey-melnychuk:
Great work, thank you! I like the decoupling of validation that allows reasonably easy testing.

One nitpick though: we're trying to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) approach. I'd imagine the commit message (and maybe a PR title to avoid confusion) along the lines of `feat(config): add validation`.

ICavlek:
I have changed the title of the PR so that it is according to the conventional commits. I guess that this is sufficient to have that message displayed with the final commit in main branch.

crates/core/src/config.rs
ICavlek: I will then add constants for each Mainnet/Sepolia Ethereum/Starknet Id and check if the provided URLs in Config get those Id's by calling *_chainId on them. If the returned Id's don't match, verification will fail. Based on that, *_blockNumber calls in check_url() are going to be replaced by *_chainId

ICavlek:
@sergey-melnychuk @LKozlowski  I have updated the PR based on the comments. Please review it again

sergey-melnychuk (approved):
LGTM,

Only one nitpick: please amend & rename commits according to conventional commits format.

ICavlek:
> conventional commits

Not sure that I follow. After finished PR, all commits should be squashed and there should be only one commit with name of the PR `feat(config): add validation`, which is according to conventional commits format. This branch would be deleted, so everything looks good to me. Please correct me if I understood something wrong

LKozlowski (approved):

LKozlowski:
> > conventional commits
> 
> Not sure that I follow. After finished PR, all commits should be squashed and there should be only one commit with name of the PR `feat(config): add validation`, which is according to conventional commits format. This branch would be deleted, so everything looks good to me. Please correct me if I understood something wrong

The idea behind that convention is that we only want meaningful commits with a bit of metadata in commit messages in the git history. In this case, we want to avoid introducing commits like "fixed clippy" into history and leave one commit that would tell us what was changed.
You are right, and as a result of your work in this pull request you should have one commit with the message "feat(config): add validation".
Of course, in the future, there might be cases where the scope of the pull request is bigger so you'll end up with more commits, but all of them should follow that convention 😄

ICavlek:
> > > conventional commits
> > 
> > 
> > Not sure that I follow. After finished PR, all commits should be squashed and there should be only one commit with name of the PR `feat(config): add validation`, which is according to conventional commits format. This branch would be deleted, so everything looks good to me. Please correct me if I understood something wrong
> 
> The idea behind that convention is that we only want meaningful commits with a bit of metadata in commit messages in the git history. In this case, we want to avoid introducing commits like "fixed clippy" into history and leave one commit that would tell us what was changed. You are right, and as a result of your work in this pull request you should have one commit with the message "feat(config): add validation". Of course, in the future, there might be cases where the scope of the pull request is bigger so you'll end up with more commits, but all of them should follow that convention 😄

Understood, I completely agree with only meaningful commits in history. I will pay attention to it in future work :)

LKozlowski:
> > > > conventional commits
> > > 
> > > 
> > > Not sure that I follow. After finished PR, all commits should be squashed and there should be only one commit with name of the PR `feat(config): add validation`, which is according to conventional commits format. This branch would be deleted, so everything looks good to me. Please correct me if I understood something wrong
> > 
> > 
> > The idea behind that convention is that we only want meaningful commits with a bit of metadata in commit messages in the git history. In this case, we want to avoid introducing commits like "fixed clippy" into history and leave one commit that would tell us what was changed. You are right, and as a result of your work in this pull request you should have one commit with the message "feat(config): add validation". Of course, in the future, there might be cases where the scope of the pull request is bigger so you'll end up with more commits, but all of them should follow that convention 😄
> 
> Understood, I completely agree with only meaningful commits in history. I will pay attention to it in future work :)

As long as you're working on something, you can have all sorts of commits as it sometimes might help to review PRs, but when you are about to merge it into the main branch then it is a time for a cleanup :)

sergey-melnychuk:
We're trying to keep the history linear (basically only fast-forward merges are allowed on the main branch), thus also try avoiding merge commits. We also don't require to squash all the commits in the PR into one, even though you can do that if you want to, so there is flexibility. If you want the PR to take only single commit - please rebase & then squash all the remaining ones.

The goal is to make small "atomic" changes that are easier to review and potentially troubleshoot. We don't have a limit on number of commits. My personal preference is to keep each change small and meaningful, even keeping `fmt` & `clippy` results (such commits are easily skipped during review and don't hurt when they sit in the branch).

PR #647, opened 2024-05-26
chore(tests): organise fixtures

sergey-melnychuk:
This PR introduces macro fixtures (that allow less boilerplate code) and an env-variable that enforces running of the tests under `experimental-api` crate.

Going further, the env-var based approach allows making sure that tests are always compiled and always run according to the environment variables being set, which potentially gives better maintainability and flexibility compared to conditional compilation based test selection.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,6 +1,5 @@
 name: check
 on:
-  workflow_dispatch:
   push:
     branches:
       - main
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -27,6 +27,7 @@ jobs:
   test-experimental:
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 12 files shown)


sergey-melnychuk:
Follow up: https://github.com/eigerco/beerus/issues/648

ICavlek (approved):
Looks good to me, nice macros. Just to double check it, integration-tests-alchemy are now run only each day at 8 o'clock, instead of every time a push on branch occurs?

LKozlowski (approved):

sergey-melnychuk:
> Looks good to me, nice macros. Just to double check it, integration-tests-alchemy are now run only each day at 8 o'clock, instead of every time a push on branch occurs?

Yes, `push` trigger disabled, `cron` one changed from 08:00 & 20:00 daily to only 08:00 daily.

PR #651, opened 2024-05-30
chore: remove hurl-based "integration tests"

sergey-melnychuk:
They do not bring any value whatsoever and test [soon-to-be] deprecated code.


ICavlek (changes requested):
Readme needs to be updated because there is an example in Getting Started section, that uses `hurl etc/rpc/starknet_getStateRoot.hurl`. TODO (in future):

- I think that this kind of simple example was nice in Readme, but for now we can be without it
- When the new api is going to be introduced from experimental, it should be mandatory to write a similar simple example in Readme that uses and shows new api's functionality

sergey-melnychuk:
> Readme needs to be updated because there is an example in Getting Started section, that uses `hurl etc/rpc/starknet_getStateRoot.hurl`. TODO (in future):
> 
>     * I think that this kind of simple example was nice in Readme, but for now we can be without it
> 
>     * When the new api is going to be introduced from experimental, it should be mandatory to write a similar simple example in Readme that uses and shows new api's functionality

Done, now the `getStateRoot` example call is defined in README without hurl.

ICavlek (approved):

PR #659, opened 2024-06-07
Fix github resource path in CONTRIBUTING.md

argentpapa:
fix github resource path:
image

ref:https://stackoverflow.com/questions/59047434/what-does-blob-in-github-correspond-to


sergey-melnychuk:
Removed the explicit link to README in https://github.com/eigerco/beerus/pull/661

PR #673, opened 2024-06-11
feat: proof validation implemented in experimental-api

### LINKED ISSUE #619: feature(experimental): add proof validation
https://github.com/eigerco/beerus/blob/96aceb2140c56361136c439d1f8acd269a2e0ce7/crates/experimental-api/src/rpc.rs#L293

ICavlek:
Proof validation exists in `crates::core::storage_proofs`. Due to new requirements, it has been also implemented for the new experimental-api. Branch workflow is as follows:

- Creation of unit tests in the` crates::core::storage_proofs` to check the original implementation
- Migration of implementation to `crates::core::experimental_api::gen`
- Migration of unit tests to validate that it has the same functionality
- Code cleanup

Things to be checked with more details:

- Unit tests that have been created - does something need to be added or changed
- `jsonrpc::error` - code value for the errors is it correctly done (possible feature `impl From<FromStrError>` to kick out boilerplate `map_err()?` or different implementation with the current scope)
- usage of `starknet::FieldElemen`t due to pedersen and poseidon hash - inside `parse_proof()` conversion from `gen::Felt` to `starknet::FieldElement` so that pedersen and poseidon hash functions could be called (additional usage of `felt_to_bits()` and `felt_from_bits()` from `crates::core::utils`)
- Beerus Client added to experimental-api bin - necessary to extract L1 block_num and state_root from node 

(Closes #619)

### DIFF
--- a/crates/experimental-api/src/gen.rs
+++ b/crates/experimental-api/src/gen.rs
@@ -9701,5 +9701,345 @@ pub mod gen {
             }
         }
     }
+
+    #[cfg(test)]
+    mod tests {
+        use crate::gen::{
+            ContractData, Felt, GetProofResult, Node, StorageKey,
+        };
+
+        #[test]
+        fn valid_one_level_parse_proof() {
+            let key = "0x0341c1bdfd89f69748aa00b5742b03adbffd79b8e80cab5c50d91cd8c2a79be1".to_string();
+            let value = Felt::try_new("0x47616d65206f66204c69666520546f6b656e")
+                .unwrap();
+            let edge_node_string = r#"[{
+                "edge": {
+                    "child": "0x47616d65206f66204c69666520546f6b656e",
+                    "path": {
+                        "len": 231,
+                        "value": "0x3dfd89f69748aa00b5742b03adbffd79b8e80cab5c50d91cd8c2a79be1"
+                    }
+                }
+            }]"#;
+            let mut proof: Vec<Node> =
+                serde_json::from_str(&edge_node_string).unwrap();
+            let ret_val = GetProofResult::parse_proof(key, value, &mut proof);
+
+            assert!(ret_val.is_some());
+            let ret_val = ret_val.unwrap();
+            assert_eq!(
+                ret_val.0,
+                "0x1e224db31dfb3e1b8c95670a12f1903d4a32ac7bb83f4b209029e14155bbca9"
+            );
+        }
+
+        #[test]
+        fn valid_five_level_parse_proof() {
+            let key = "0x0341c1bdfd89f69748aa00b5742b03adbffd79b8e80cab5c50d91cd8c2a79be1".to_string();
+            let value = Felt::try_new("0x47616d65206f66204c69666520546f6b656e")
+                .unwrap();
+            let proof_string = r#"[
+            {
+                "binary": {
+                    "left": "0x46e82293b0564764a071f1aa4488aa7577b1b5bb2e898321f8536d5593d371d",
+                    "right": "0x58adcf6ea8b96992aa316e2f092f2480ca406c3630fe97573046a32900745b5"
+                }
+            },
+            {
+                "binary": {
+                    "left": "0x716e211c75f4c0e14dbe46c361812b0129abd061b63faf91ad5569bf22b785c",
+                    "right": "0x3729d9699d4410223e413f3b3aa91a043d94242f888188036e6ea25b6962041"
+                }
+            },
+            {
+                "edge": {
+                    "child": "0x6281e42b5941ae1a77ea03836aad1190097f72e1a1ed534fae2e00b4118f504",
+                    "path": {
+                        "len": 1,
+                        "value": "0x1"
+                    }
+                }
+            },
+            {
+                "binary": {
+                    "left": "0x3e3800516f62800ef6491b1cb1915b3353026ea6a6afcf35e8d4c54e35b04ea",
+                    "right": "0x1e224db31dfb3e1b8c95670a12f1903d4a32ac7bb83f4b209029e14155bbca9"
+                }
+            },
+            {
+                "edge": {
+                    "child": "0x47616d65206f66204c69666520546f6b656e",
+                    "path": {
+                        "len": 231,
+                        "value": "0x3dfd89f69748aa00b5742b03adbffd79b8e80cab5c50d91cd8c2a79be1"
+                    }
+                }
+            }]"#;
+            let mut proof: Vec<Node> =
+                serde_json::from_str(&proof_string).unwrap();
+            let ret_val = GetProofResult::parse_proof(key, value, &mut proof);
+
+            assert!(ret_val.is_some());
+            let ret_val = ret_val.unwrap();
+            assert_eq!(
+                ret_val.0,
+                "0x6cc50a732b4256f7b642348e19bd1a8bee7ac76bed3fcee3bc34309538c00c6"
+            );
+        }
+
+        #[test]
+        fn invalid_one_level_parse_proof() {
+            let key = "0xabc".to_string();
+            let value = Felt::try_new("0xdef").unwrap();
+            let mut proof: Vec<Node> = serde_json::from_str(
+                r#"[{
+                "edge": {
+                    "child": "0xfaa",
+                    "path": {
+                        "len": 1,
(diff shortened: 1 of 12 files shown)


crates/experimental-api/src/gen.rs
```diff
@@ -102,7 +112,7 @@ pub mod gen {
 
     #[derive(Clone, Debug, Deserialize, Serialize)]
     #[serde(try_from = "i64")]
-    pub struct BlockNumber(i64);
+    pub struct BlockNumber(pub i64);
```
sergey-melnychuk, line 115: Now `BlockNumer` can be negative :)

crates/experimental-api/src/gen.rs
```diff
@@ -2162,6 +2185,312 @@ pub mod gen {
         pub state_commitment: Option<Felt>,
     }
 
+    impl GetProofResult {
```
sergey-melnychuk, line 2188: Please move out the implementation out of `gen.rs`, it is only for DTOs and relevant validation.

crates/experimental-api/src/gen.rs
```diff
@@ -2137,7 +2160,7 @@ pub mod gen {
         pub transaction_hash: Option<Felt>,
     }
 
-    #[derive(Clone, Debug, Deserialize, Serialize)]
+    #[derive(Clone, Debug, Deserialize, Serialize, Default)]
```
sergey-melnychuk, line 2163: What exactly does `Default` for `ContractData` mean? What such a contract can do?

crates/experimental-api/src/gen.rs
```diff
@@ -1659,6 +1676,12 @@ pub mod gen {
                 &self.0
             }
         }
+
+        impl From<StorageKey> for String {
+            fn from(value: StorageKey) -> Self {
+                value.0
```
sergey-melnychuk, line 1682: There is already `impl AsRef` doing exactly this

crates/experimental-api/src/gen.rs
```diff
@@ -996,6 +1007,12 @@ pub mod gen {
                 &self.0
             }
         }
+
+        impl From<Address> for Felt {
+            fn from(value: Address) -> Self {
+                value.0
```
sergey-melnychuk, line 1013: There is already `AsRef`

crates/experimental-api/src/gen.rs
```diff
@@ -954,13 +964,14 @@ pub mod gen {
         pub unit: PriceUnit,
     }
 
-    #[derive(Clone, Debug, Deserialize, Serialize)]
+    #[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq)]
```
sergey-melnychuk, line 967: What is a "default Felt"? Is it zero? Empty sting? It is unclear, thus a user needs to explicitly state what they need.

crates/experimental-api/src/gen.rs
```diff
@@ -9,10 +9,20 @@ pub mod gen {
...
+
+    use beerus_core::storage_proofs::types::Direction;
+    use beerus_core::utils::{felt_from_bits, felt_to_bits};
 
     #[derive(Clone, Debug, Deserialize, Serialize)]
     pub struct Address(pub Felt);
 
+    impl From<Address> for String {
+        fn from(value: Address) -> String {
+            value.0 .0
```
sergey-melnychuk, line 22: `.as_ref().as_ref()` would work instead

crates/experimental-api/src/gen.rs
```diff
@@ -9,10 +9,20 @@ pub mod gen {
     use serde_json::Value;
 
     use iamgroot::jsonrpc;
+    use starknet_crypto::{pedersen_hash, poseidon_hash_many, FieldElement};
```
sergey-melnychuk, line 12: In general `gen.rs` contains only generated code and is not expected to be updater directly.

If something is missing, it is added to the codegen tool, so on each change of the spec painful merges can be avoided.

I will put this into README/CONTRIBUTING.

crates/experimental-api/src/rpc.rs
```diff
@@ -291,6 +299,28 @@ impl gen::Rpc for Context {
         key: StorageKey,
         block_id: BlockId,
     ) -> std::result::Result<Felt, jsonrpc::Error> {
+        let l1_block_num = self.node.read().await.l1_block_number as i64;
```
sergey-melnychuk, line 302: This block seems like a good candidate for a method/function extraction (and maybe relevant unit-testing if it makes sense).

sergey-melnychuk (changes requested):

crates/experimental-api/src/gen.rs
ICavlek: Changed back to private

crates/experimental-api/src/gen.rs
ICavlek: Migrated to newly created proof_validation module

crates/experimental-api/src/gen.rs
ICavlek: Removed Default

crates/experimental-api/src/gen.rs
ICavlek: Removed

crates/experimental-api/src/gen.rs
ICavlek: Removed

crates/experimental-api/src/gen.rs
ICavlek: Removed Default

crates/experimental-api/src/gen.rs
ICavlek: Removed

crates/experimental-api/src/gen.rs
ICavlek: Removed

crates/experimental-api/src/rpc.rs
ICavlek: Migrated to `impl Context`. Added proper unit tests

crates/experimental-api/tests/rpc.rs
```diff
@@ -252,8 +252,7 @@ async fn test_getStorageAt() -> Result<(), Error> {
         "0x0341c1bdfd89f69748aa00b5742b03adbffd79b8e80cab5c50d91cd8c2a79be1",
     )?;
 
-    let block_id =
-        BlockId::BlockNumber { block_number: BlockNumber::try_new(600612)? };
```
sergey-melnychuk, line 256: why this change is necessary?

crates/experimental-api/src/rpc.rs
```diff
@@ -103,6 +110,40 @@ impl IntoResponse for RpcError {
...
+                let mut block_num = l1_block_num;
+                let user_request_block_num = *block_number.as_ref();
+                if user_request_block_num < block_num {
+                    block_num = user_request_block_num;
+                }
+                Ok(BlockId::BlockNumber {
+                    block_number: BlockNumber::try_new(block_num)?,
+                })
+            }
+            gen::BlockId::BlockHash { block_hash } => {
```
sergey-melnychuk, line 133: For block hashes it might be trickier that that. Ideally we should resolve the block hash into the block number and check if that block number is acceptable.

Please create a follow up issue for that and put a comment here referencing the issue, just to avoid scope extension for this pull request.

ICavlek: Created https://github.com/eigerco/beerus/issues/675 and added to backlog of the [beerus](https://github.com/orgs/eigerco/projects/27) github project

crates/experimental-api/tests/rpc.rs
ICavlek: I have pushed the last commit with the revert back to 600612 for visualization. There is also a change in `ctx()` setup of node's `l1_state_root`. 

In `crates::experimental-api::src::proof_validation::mod.rs` there is a check `*state_commitment == parsed_global_root && global_root == parsed_global_root`. 

They didn't match because of the wrong `l1_state_root`, making the test fail and I have adapted `block_id` at first. Both tests are valid, with `Tag::Latest` and 600612, but when I was testing it on my local pc, I was constantly using `Tag::Latest` for calling rpc and it was easier for me to update the test with these values.

Issue #684, opened 2024-06-18
feat(ci): report code coverage


sergey-melnychuk:
https://github.com/taiki-e/cargo-llvm-cov
https://github.com/mozilla/grcov

jrmncos:
hey team! I would like to tackle this one

sergey-melnychuk:
Current state: 

`cargo tarpaulin --exclude-files src/gen.rs --exclude-files src/bin/* --exclude-files web/* --out html`

<img width="559" alt="Screenshot 2024-09-28 at 21 00 24" src="https://github.com/user-attachments/assets/14a4ca87-96a8-4c4b-9e10-58b06ab26091">

Unfortunately running coverage job with `tarpaulin` turns out to be too much for a Github Actions worker, so right now I am looking for a workaround.

sergey-melnychuk:
https://github.com/eigerco/beerus/pull/787#issuecomment-2431369794

PR #719, opened 2024-07-18
feat: adding source based coverage using cargo-llvm-cov

jrmncos:
Using [cargo-llvm-coverage ](https://github.com/taiki-e/cargo-llvm-cov) to generate source based coverage. 
The library is wrapper of the compiler flag `-C instrument-coverage` more details [here](https://doc.rust-lang.org/rustc/instrument-coverage.html#instrumentation-based-code-coverage)

The command `cargo llvm-cov --all-features --workspace --html --output-dir coverage` was tested locally and it generated an html file with the coverage.

I used Github page to publish the .html file, but Codecov can be used too. I need to verify is there are free plan options.

Any change or suggest will be welcome!

#684

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -83,3 +83,23 @@ jobs:
       - uses: crate-ci/typos@v1.21.0
         with:
           files: .
+
+  coverage:
+    if: ${{ !startsWith(github.head_ref, 'dependabot/') }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        run: rustup update stable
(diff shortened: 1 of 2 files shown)


jrmncos:
btw the integration with Github pages is not fully tested. I couldn't from my forked repo, cause I didn't have permissions. I published the PR mainly to discuss about it. Thanks!

jrmncos:
@sergey-melnychuk hey Sergei how it's going? I used Github pages because I'm not so sure if codecov is free. I can research about that and implemented. What do you think?

sergey-melnychuk:
https://github.com/eigerco/beerus/pull/787#issuecomment-2431369794

PR #753, opened 2024-08-27
feat: upgrade to 0.7.1 rpc spec

sergey-melnychuk:
0.7.0: https://github.com/starkware-libs/starknet-specs/tree/v0.7.0
0.7.1: https://github.com/starkware-libs/starknet-specs/compare/v0.7.0...v0.7.1

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BEERUS_TEST_RUN: 1
-      BEERUS_TEST_STARKNET_URL: ${{ secrets.STARKNET_RPC_0_6_0_MAINNET }}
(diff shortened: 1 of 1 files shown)


ICavlek (changes requested):
I have tried it out with declaring V3 accounts on Katana with starkli and the newly added skip flag. It worked. Because of the failing tests, things that need to be updated:

- tests/rpc.rs `test_specVersion` is failing - I guess that it still calls 0.6 rpc api because it returns 0.6.0. Changing it to {https.//starknet-mainnet}.../rpc/v0_7/{API_KEY} in the secrets part of the enviroment should do it
- When first bullet is fixed, I get in tests/exe.rs failing `test_call_regular_contract_class`. I guess that due to rpc changes, this also needs to be adapted
- Not essential, but I think that it would be nice to also update README.md and state that we have migrated to rpc 0.7.1

ICavlek (approved):
LGTM

PR #767, opened 2024-09-10
fix(exe/call): get proof for correct block

sergey-melnychuk:
Closes https://github.com/eigerco/beerus/issues/768


ICavlek (approved):
LGTM

PR #770, opened 2024-09-13
feat!(config): drop `network` config property

xep-c-6yrpa:
The main goal of this PR is to suggest rather radical change in configuration of Beerus: get rid of the `network` config property - it is completely redundant, because RPC URL for Ethereum and Starknet identify the used chain. In such case the detection of the network is very simple and allows no space for misconfiguration: only if both Etherem and Starknet chain IDs match (for either Ethereum or Sepolia), then network is detected correctly. In all other cases (unless 'testing' feature is enabled, then support of Katana chain was added explicitly) the error is returned. 

Other less significant suggestions:
- drop JSON configuration support and leave only TOML
- fail when invalid config properties were provided (now it falls back to the defaults for some reason)
- log updated state only if it actually differs from the current one
- update and cleanup relevant test coverage
- sheck expected RPC version on the client (currently for some reason it is only done for server in the `beerus` binary)

I know that suggested changes are rather significant, but I believe it still makes sense to consider them, as they making Beerus more consistent and more usable. Ready to discuss in anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,55 +153,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anstream"
-version = "0.6.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
-dependencies = [
- "anstyle",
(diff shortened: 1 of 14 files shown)


sergey-melnychuk (approved):
LGTM

PR #793, opened 2024-09-30
feat: add support in test environment for sepolia network

ICavlek:
~~Test that calls declare and deploy with account data on sepolia network is introduced with this PR. It simulates a way on how curl could directly call beerus that is connected with sepolia network. Since this transaction needs gas and prefunded wallets, it has been decided to have this test on sepolia with an already deployed same account. Results of the test should be errors.
Test in a nutshell:~~

~~- Chain id and spec verification~~
~~- *EDIT* Get Class results in ok value because the account already exists on sepolia~~
~~- Account declaration results in error~~
~~- Account deployment results in error~~

Things that were updated based on test requirements:

- Migrated values from constants.rs to dummies.rs because the values inside are not exactly constants
- Migrated constants from account_katana.rs to constants.rs so that they could be reused
- Made constants from config.rs public so that they could be reused
- Refactored `setup!()` macro to support also sepolia network

EDIT: Updated PR name because of the update on issue


.github/workflows/check.yml
```diff
@@ -28,7 +28,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BEERUS_TEST_RUN: 1
-      BEERUS_TEST_STARKNET_URL: https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0_7/${{ secrets.ALCHEMY_KEY }}
```
sergey-melnychuk, line 31: Test context is currently network-agnostic, why not just keep `BEERUS_TEST_STARKNET_URL` and set it to the network-specific URL for specific test run?

tests/common/ctx.rs
```diff
@@ -40,15 +63,22 @@ pub async fn ctx() -> Option<Context> {
...
             .ok()
             .map(|value| &value == "1")
             .unwrap_or_default();
         if !run {
             return Ok(());
         }
-        if let Some(ctx) = common::ctx::ctx().await {
+        let context = match $e {
+            "sepolia" => common::ctx::ctx_sepolia().await,
+            _ => common::ctx::ctx_mainnet().await,
```
sergey-melnychuk, line 79: This allows for funny and non-obvious things like `setup!("lol")` to run against mainnet. Making it more strict maybe?

tests/rpc.rs
```diff
@@ -436,3 +444,53 @@ async fn test_getClassHashAt() -> Result<(), Error> {
     );
     Ok(())
 }
+
+#[tokio::test]
+async fn test_declare_deploy_account_already_existing() -> Result<(), Error> {
+    let ctx = setup!("sepolia");
```
sergey-melnychuk, line 450: if both DECLARE and DEPLOY_ACCOUNT transactions are expected to fail, what is the point of running it on sepolia only?

.github/workflows/check.yml
sergey-melnychuk: Oh, I see - the subset of tests is expected to run against testnet.

tests/rpc.rs
ICavlek: Because this account already exists on sepolia, it does not on mainnet. Calling `addDeclareTransaction` and `addDeployAccountTransaction` would start the process of adding new account to the mainnet which I think is not suitable for a test case at this point. Maybe some time when we will have budget of starks for mainnet, it could be taken into account.

ICavlek: Forgot to write in description that also `getClass` is called that verifies that account exists on sepolia. This way if somehow setup comes to mainnet instead of sepolia in this specific test case, this should fail first before any bamboozle declare/deploy calls would happen.

tests/common/ctx.rs
ICavlek: Updated

tests/rpc.rs
sergey-melnychuk: > adding new account to the mainnet

> is not suitable for a test case at this point

Not trying to be funny here, just thinking what this test actually tests.

The way I see it, it tests that it is impossible to declare & deploy an existing account. If the test fails, it means that somehow the existing account go declared & deployed again. So it actually tests Starknet.

What we neet to test is that the it is possible (using owr code) to deploy a new account. We might want to run this test on demand (idk, once a month), on provided input (secret key). Then if such test fails, it means that probably something is wrong with our code, and we can verify that (e.g. run A/B tests with/without Beerus as a proxy). Such test would make sense to run on testnet (to fee in SepoliaETH not it real ETH).

tests/common/ctx.rs
```diff
@@ -40,15 +63,23 @@ pub async fn ctx() -> Option<Context> {
...
             .map(|value| &value == "1")
             .unwrap_or_default();
         if !run {
             return Ok(());
         }
-        if let Some(ctx) = common::ctx::ctx().await {
+        let context = match $e {
+            "sepolia" => common::ctx::ctx_sepolia().await,
+            "mainnet" => common::ctx::ctx_mainnet().await,
+            _ => panic!("Invalid network setup. Has to be mainnet or sepolia."),
```
sergey-melnychuk, line 80:
```suggestion
            unknown => panic!("Unknown network: {unknown}. Supported networks: mainnet, sepolia"),
```

tests/rpc.rs
sergey-melnychuk: > Since this transaction needs gas and prefunded wallets, it has been decided to have this test on sepolia with an already deployed same account.

Prefunded wallet on Sepolia is not a problem, rather a pre-requisite for the test. So decision to run the test against deployed account renders this test pointless, unfortunately.

README.md
```diff
@@ -147,7 +147,8 @@ cargo build --release
 cargo test
 
 ## Run integration tests against live endpoint
-export BEERUS_TEST_STARKNET_URL=https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0_7/${ALCHEMY_API_KEY}
+export STARKNET_MAINNET_URL=https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0_7/${ALCHEMY_API_KEY}
+export STARKNET_SEPOLIA_URL=https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0_7/${ALCHEMY_API_KEY}
```
sergey-melnychuk, line 151:
```suggestion
export STARKNET_SEPOLIA_URL=https://starknet-sepolia.g.alchemy.com/starknet/version/rpc/v0_7/${ALCHEMY_API_KEY}
```

ICavlek:
@sergey-melnychuk I have removed unnecessary test. Still, I would like to use this PR to close this issue: https://github.com/eigerco/beerus/issues/802
Additionally, I was able to create tests which would solve calling existing contracts: https://github.com/eigerco/beerus/issues/696
Contracts that are called are on sepolia, and that PR would rely on this PR which adds possibility to connect beerus with sepolia in test environment.
Preview of the tests are in branch ic/call_existing_contract.

sergey-melnychuk (approved):

Issue #804, opened 2024-10-03
feat: starkli support for declaring and deploying account


ICavlek:
Since beerus is a middleware, it is necessary to support starkli that is user interface for interacting with starknet. It has to be investigated and decided how it is going to be: from env via .sh script, as docker image or it will be imported as a library directly in the test. Following features should be supported:

- `starkli signer` - for creating public/private key
- `starkli declare` - for declaring account contract
- `starkli invoke eth transfer` - for funding new account
- `starkli deploy` - for deploying account contract

PR #813, opened 2024-10-14
test: scarb library with template account contract

ICavlek:
In scope of this PR, scarb library has been introduced in test environment. Scarb is a tool that is used for compilation of smart contracts within the starknet ecosystem. This is marked as an initial step for actual account deployment within test cases.
General overview of the introduced changes:

- Added two test cases, one for sepolia and one for katana
- Added template account contract
- Possibility to create and compile multiple account contracts

Additional notable changes:

- Refactor of test module
- Added TODOs for additional work

Since the general overview of the test case can be observed with the changes in this PR and future TODOs, please consider the structure and the workflow of the test case within this PR.


LKozlowski (approved):
LGTM

To make sure I understand it correctly: We use scarb from within tests to compile account's smart contract then deploy it and test it?

ICavlek:
> LGTM
> 
> To make sure I understand it correctly: We use scarb from within tests to compile account's smart contract then deploy it and test it?

Yes, scarb is used within the executor. I have created a template where multiple accounts will be deployed on Katana before using the framework on Sepolia. Since the compilation is a slow process (~30 secs in test), it is pushed in separate threads based on the amount of newly deployed accounts.
After that, with starkli  introduced also as lib ([issue](https://github.com/eigerco/beerus/issues/804) that I am currently working on) we will be able to declare and deploy it on network

tests/rpc.rs
```diff
@@ -532,3 +511,17 @@ async fn account_call() -> Result<(), Error> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn deploy_new_account_on_sepolia() -> Result<(), Error> {
+    // TODO #807
```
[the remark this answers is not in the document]
ICavlek: I agree. I have thought it through and came to three obstacles :

1. Make changes within template contract so that new account is deployed - simple id update
2. Commit changes within github workflows to have automated test

Last thing that comes to my mind, is that with every new account we get `key.json` and `account.json` files which are needed for `declare` and `deploy` on Sepolia network. This gives two options which I need help with decision:

1. We can easily delete them when test ends and leave accounts in the deserts of the blockchain without the ability to access them
2. We can store all these values in some additional folder so that we can access them later

I guess that storing them would be a better way and committing them in some folder, but having in mind that `key.json` and `account.json` might not be the best things to have publicly exposed.

tests/starknet/executor.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        Ok(())
+    }
+
+    fn compile(&self) -> Result<(), Error> {
+        let mut vec_of_threads = Vec::with_capacity(self.accounts.len());
+
+        for account in self.accounts.iter() {
+            let path = account.clone() + "/Scarb.toml";
+            let compilation = thread::spawn(move || -> Result<(), Error> {
```
sergey-melnychuk, line 91: Why go with raw threads? We already have `tokio` in dev-dependencies anyway.

tests/starknet/executor.rs
```diff
@@ -0,0 +1,128 @@
...
+                let compiler = Compiler::new(&path)?;
+                compiler.compile()
+            });
+            vec_of_threads.push(compilation);
+        }
+        for (i, thread) in vec_of_threads.into_iter().enumerate() {
+            let compilation = thread.join();
```
sergey-melnychuk, line 98: With tokio it seems like a use-case for [JoinSet](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html)

tests/starknet/executor.rs
```diff
@@ -0,0 +1,128 @@
...
+use super::scarb::Compiler;
+
+#[allow(dead_code)]
+pub struct Executor {
+    accounts: Vec<String>,
+}
+
+#[allow(dead_code)]
+impl Executor {
+    pub fn new(num: usize) -> Result<Self, Error> {
```
sergey-melnychuk, line 15: Why embed threading in the executor? It seems much easier to get an executor for a specific `id` and let client code decide if any number of them is necessary. Right now I don't really see the use case (please correct me if I'm wrong) for number of contracts deployed being anyhting that one :)

tests/starknet/executor.rs
```diff
@@ -0,0 +1,128 @@
...
+        // #804 #805 fund accounts from pre-funded account
+        // #804 deploy accounts
+        // #806 iterate through class hashes and call getClass to see if they are verified
+        Ok(())
+    }
+
+    fn update_account(&self, path: &str) -> Result<(), Error> {
+        let lib_path = path.to_owned() + "/src/lib.cairo";
+        let account_old = fs::read_to_string(lib_path.clone())?;
+        let re = Regex::new(r"self.id.write\((?<number>\d+)\);")?;
```
sergey-melnychuk, line 44: It might be easier to extract a constant in the `lib.cairo` and patch just it, not the actual code. Or add a placeholder instead of zero, as the contract is not expected to be runnable/deployable outside the executor. Then simple `replace` should be enough.

tests/starknet/executor.rs
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl Drop for Executor {
+    fn drop(&mut self) {
+        let dir = self.accounts[0].clone() + "/target";
```
sergey-melnychuk, line 113: Just as an alternative: copy the contract into `./target/account-${SUFFIX or timestamp}`, do the manipulations, and just leave it there. If any line here fails (rather unlikey though) it leaves working directory messed up.

tests/starknet/scarb.rs
```diff
@@ -0,0 +1,43 @@
...
+    toml: PathBuf,
+    opts: CompileOpts,
+    packages: Vec<PackageId>,
+}
+
+#[allow(dead_code)]
+impl Compiler {
+    pub fn new(toml: &str) -> Result<Self, Error> {
+        let toml_absolute = canonicalize(PathBuf::from(toml))?;
+        let opts = CompileOpts {
```
sergey-melnychuk, line 20: If `opts` and `packaged` are hard-coded and never changed, why keep them in `Compiler` as fields and not just use them in `compile`? Then whole `compile` seems to be a single function that takes `toml: &str` and returns `Result`.

tests/starknet/executor.rs
ICavlek: Since the compilation is a blocking call and it goes into a separate thread, I started to think one step ahead. I wanted to make Executor reusable with simple API where the amount of contracts is going to be input and the rest of it will be done implicitly, including environment setup of followup files such as account.json, compilation by scarb, usage of starkli for declare and deploy, etc..
Additionally, before running this test case on live network, I plan to test it on Katana (hence, two tests, one for Katana and one for Sepolia). Since iterating multiple accounts compilation would take some time on a test case, with simple API in mind, this is the approach that I have decided on.
I agree with you, in the end we want to have only one contract that will be deployed. With this PR, I wanted to put everything on table, show what design questions came to my mind during the work and what we are dealing with. After the review, I could easily kick something out, if you thought that something was being overengineered.
To summarize, next steps that will be done, I will extract compilation outside of the executor and in the test case with Katana where we will deploy multiple accounts, compilation of the accounts is going to be called explicitly, directly inside the test case.

tests/starknet/executor.rs
ICavlek: I will go with the constant and not a placeholder, just for us so that we have stored information in the contract which was the last value used.
I wanted to have it as part of a contract so that maybe, some time in the future, we might have additional test cases that will involve interaction of this accounts that we have deployed, i.e. some marketplace.
Therefore, I will go with a constant, if it will be possible to call account and get the value of the constant. I was sure about this id, with creating function id(). I have to investigate whether this is the case for constants.

tests/starknet/executor.rs
ICavlek: @sergey-melnychuk Result of the changes is that `executor.rs` has been removed and that `coordinator.rs` is introduced. Coordinator's task is to store all paths and copy/update contract that is being worked on. I have removed support for multiple contract deployment and inside the test case explicit workflow can be checked. The one in `rpc.rs` that will be used to deploy accounts on sepolia is commented out, for now. I will uncomment it when we will test whole workflow on Katana.

tests/starknet/executor.rs
ICavlek: @sergey-melnychuk Updated template contract, using placeholder <ID> and replacing it with timestamp that has format YMDHMS (e.g. 20241023140955).

ICavlek:
@sergey-melnychuk @LKozlowski I have updated PR with significant changes and I would like you to have another look at it. Based on the comments, I have reworked it in a way that I have removed `executor.rs` and added `coordinator.rs` which will only handle the environment of the contracts.

One additional thing that is not directly connected to this PR, clippy fails in `gen.rs` (see workflows on commits). I have updated it here so that it can be observed, but I guess that this will be done in a different PR. I will update it here when it is done.

tests/account_katana.rs
```diff
@@ -41,6 +48,24 @@ async fn declare_deploy_account_v2() {
...
 
+#[tokio::test]
+async fn deploy_new_account_on_katana() -> Result<(), Error> {
+    let _katana = Katana::init("http://127.0.0.1:0").await?;
+    let coordinator = Coordinator::new(TestMode::Katana);
+    coordinator.copy_template_to_target()?;
+    coordinator.update_account()?;
+    let compiler = Compiler::new(&coordinator.target_scarb())?;
+    compiler.compile().await?;
+    // TODO
```
sergey-melnychuk, line 59: Seems like a good candidate for extracting into a separate issue.

tests/rpc.rs
```diff
@@ -532,3 +510,26 @@ async fn account_call() -> Result<(), Error> {
...
```
sergey-melnychuk, line 516: Maybe extract it into another issue and separate PR

tests/starknet/scarb.rs
sergey-melnychuk: 20 lines of code is not a hure function. Having a single function is better that having a pointless struct, don't you think?

tests/starknet/scarb.rs
```diff
@@ -0,0 +1,57 @@
...
+            tokio::task::spawn_blocking(move || -> Result<(), Error> {
+                self.run_compilation()
+            });
+        match compilation.await {
+            Ok(val) => Ok(val?),
+            Err(e) => Err(anyhow!(
+                "Error during thread execution. Original error message: {:#?}",
+                e,
+            )),
+        }
```
sergey-melnychuk, line 49:
```suggestion
        tokio::task::spawn_blocking(move || {
            let result = self.run_compilation()?;
            Ok(result)
        })?
        // might require extra error variant
```

tests/starknet/coordinator.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    pub fn target_src(&self) -> String {
+        self.target.clone() + "/src"
+    }
+}
+
+impl Drop for Coordinator {
+    fn drop(&mut self) {
+        match self.mode {
+            TestMode::Katana => {
```
sergey-melnychuk, line 73: Why do we neet this explicit clean up? Why only with Katana but not Sepolia?

tests/starknet/coordinator.rs
```diff
@@ -0,0 +1,82 @@
...
+        let target = "./target/account-".to_string() + &id;
+        Self {
+            id,
+            mode,
+            source: "./tests/starknet/contract/account".to_string(),
+            target,
+        }
+    }
+
+    pub fn copy_template_to_target(&self) -> Result<(), Error> {
```
sergey-melnychuk, line 34: The problem with such design is that `Coordinator` instance is created in the invalid state, and is unusable until methods are called in specific order. Why not extract them as a functions and use in `Coordinator`? Then whole `Coordinator` can be substituted with a simple function (idk, maybe along the lines of `prepare_account`) which is impossible to misuse.

tests/rpc.rs
```diff
@@ -1,40 +1,18 @@
...
+    TxnExecutionStatus, TxnHash, TxnReceipt, TxnReceiptWithBlockInfo,
+    TxnStatus,
 };
 
 mod common;
+mod starknet;
 
 use common::err::Error;
 
-#[tokio::test]
```
sergey-melnychuk, line 18: Why remove these test cases?

sergey-melnychuk (changes requested):
Left a few comments about unnecessary complications if a very simple workflows (introducing structs with hard-coded unchangeable properties, when all the process can be described as a composition of a simple functions).

tests/starknet/executor.rs
sergey-melnychuk: Please don't leave chunks of commented code in the pull-request, it is a hure code smell.

Extract it into separete issues/PRs if necessary.

tests/rpc.rs
```diff
@@ -1,40 +1,18 @@
...
-    let ctx = setup!();
-
-    let ret = ctx.client.specVersion().await?;
-    assert_eq!(ret, "0.7.1");
-    Ok(())
-}
-
-#[tokio::test]
-#[allow(non_snake_case)]
-async fn test_chainId() -> Result<(), Error> {
```
sergey-melnychuk, line 30: Why remove this test case?

tests/starknet/coordinator.rs
```diff
@@ -0,0 +1,82 @@
...
+#[allow(dead_code)]
+impl Coordinator {
+    pub fn new(mode: TestMode) -> Self {
+        let now = chrono::offset::Local::now();
+        let id = now.format("%Y%m%y%H%M%S").to_string();
+        let target = "./target/account-".to_string() + &id;
+        Self {
+            id,
+            mode,
+            source: "./tests/starknet/contract/account".to_string(),
```
sergey-melnychuk, line 29: Why hard-code it as a property when it is effectively a constant?

tests/starknet/coordinator.rs
sergey-melnychuk: If cleanup is unified, we don't need a distinction between networks as well as `TetsMode` enum.

tests/starknet/scarb.rs
```diff
@@ -0,0 +1,57 @@
...
+        let packages = vec![PackageId::new(
+            PackageName::new("account"),
+            Version::new(0, 1, 0),
+            SourceId::for_path(toml_absolute.to_str().unwrap().into())?,
+        )];
+
+        Ok(Compiler { toml: toml_absolute, opts, packages })
+    }
+
+    pub async fn compile(self) -> Result<(), Error> {
```
sergey-melnychuk, line 38: Why not just make the method blocking and let client code decide whether it needs to be run with `spawn_blocking` or not?

Remember, "premature optimization is the root of all evil."

src/gen.rs
```diff
@@ -1118,7 +1118,7 @@ pub mod gen {
     #[derive(Clone, Debug, Deserialize, Serialize)]
     #[serde(untagged)]
     pub enum InvokeTxnTraceExecuteInvocation {
-        FunctionInvocation(FunctionInvocation),
+        FunctionInvocation(Box<FunctionInvocation>),
```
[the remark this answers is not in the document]
ICavlek: I know that it is generated code and that it worked fine, I have just made it visible for you so that you can observe what clippy was complaining about. I didn't know that it was already addressed in a different PR, not until I started reviewing it, after this push. I will rebase it.

tests/starknet/coordinator.rs
ICavlek: My process of thought was that if we will be deploying accounts on sepolia with monthly scheduled tests, the ones deployed for Katana will be deleted and the ones that will be running on sepolia, could be stored and reused for some further test cases (i.e. some marketplace). Additionally, in the account-{timestamp} folder, there will be files such as `key.json` and `account.json` which will be generated by `starkli` in next issue.

I have went through all of your comments and this one also goes under:
>Remember, "premature optimization is the root of all evil."

tests/rpc.rs
ICavlek: I guess formatting issue O.o. Returning it

tests/rpc.rs
ICavlek: I guess formatting issue, same as one above. Returning it

ICavlek:
@sergey-melnychuk @LKozlowski Updated PR based on comments: 
- Recovered two accidentally deleted tests
- Removed two template tests that will be added in scope of separate PRs
- Removed `Compiler` class and reduced it to single function
- Removed `coordinator.rs`, it's `Coordinator` class and reduced it to single function in `utils.rs`

sergey-melnychuk (approved):

PR #840, opened 2024-11-20
bugfix: has to be a list cron type

ICavlek:
Based on the error in actions:

`schedule accepts a list of one or more maps with the cron key set`

This should fix it and run test at 10 o'clock on each first of month

### DIFF
--- a/.github/workflows/account_deployment.yml
+++ b/.github/workflows/account_deployment.yml
@@ -1,7 +1,7 @@
 name: account-deployment
 on:
   schedule:
-      interval: "monthly"
+    - cron: 0 10 1 * *
 
 env:
   RUSTFLAGS: -D warnings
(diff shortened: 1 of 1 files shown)


.github/workflows/account_deployment.yml
```diff
@@ -1,7 +1,7 @@
 name: account-deployment
 on:
   schedule:
-      interval: "monthly"
+    - cron: 0 10 1 * *
```
sergey-melnychuk, line 4: I think you need to quote the cron-expression

sergey-melnychuk (approved):

.github/workflows/account_deployment.yml
ICavlek: You are right, tnx

